PR #20, opened 2015-01-15
Update to use buildpack-deps:*-micro as the base

### LINKED ISSUE #17: use base library/buildpack-deps
It would be nice if the Java images could use `library/buildpack-deps` as the base image. Most of the language stacks use this base image, and it would be nice to have consistency. This image also comes with some useful utilities, such as git, which can be helpful when using the image as part of a build process.

tianon:
Fixes #17

### DIFF
--- a/openjdk-6-jdk/Dockerfile
+++ b/openjdk-6-jdk/Dockerfile
@@ -1,14 +1,16 @@
-FROM debian:wheezy
+FROM buildpack-deps:wheezy-micro
 
 # A few problems with compiling Java from source:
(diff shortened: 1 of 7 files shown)


yosifkit:
you should to drop `curl` and `wget` since they are in micro. :smile:

tianon:
Sssh, don't judge me.

tianon:
Doing a build test now to make sure I didn't break anything in my refactoring.

tianon:
:+1:

yosifkit:
LGTM

PR #23, opened 2015-01-21
Apply ca-certificates-java workaround for now (https://bugs.debian.org/775775)

### LINKED ISSUE #19: ubuntu java package has broken cacerts
See https://bugs.launchpad.net/ubuntu/+source/ca-certificates-java/+bug/1396760

At this time, the latest image id aec8201c9d63 has no /etc/ssl/certs/java/cacerts file:

    $ docker run --rm -it java:openjdk-8-jdk ls -al /etc/ssl/certs/java/cacerts
    ...
    ls: cannot access /etc/ssl/certs/java/cacerts: No such file or directory

Note that this is not broken in java 7
    $ docker run --rm -it java:openjdk-7-jdk ls -al /etc/ssl/certs/java/cacerts
    ...
    -rw-r--r-- 1 root root 206373 Jan  1 07:13 /etc/ssl/certs/java/cacerts

This means that https cannot be used in these images (among other issues), which breaks downstream images like maven:3.2-jdk-8 (which is where I discovered the issue).

A workaround (from the linked issue) is to add command to the build, after the apt-get command:

    /var/lib/dpkg/info/ca-certificates-java.postinst configure

Note that `sudo` is left off, as the image doesn't have sudo, and runs as root anyway.

tianon:
Fixes #19

### DIFF
--- a/openjdk-8-jdk/Dockerfile
+++ b/openjdk-8-jdk/Dockerfile
@@ -10,7 +10,14 @@ RUN apt-get update && apt-get install -y unzip && rm -rf /var/lib/apt/lists/*
 ENV JAVA_VERSION 8u40~b09
 ENV JAVA_DEBIAN_VERSION 8u40~b09-1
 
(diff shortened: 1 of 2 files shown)


yosifkit:
LGTM

PR #33, opened 2015-04-08
Clean downloaded package files cache

pataquets:


yosifkit:
The debian base images have this built in to `apt-get update` ([contrib/mkimage/debootstrap](https://github.com/docker/docker/blob/master/contrib/mkimage/debootstrap#L98)):smile:.

PR #34, opened 2015-04-14
Make file.encoding default to UTF-8

### LINKED ISSUE #32: UTF8 encoding by default?
I've run into this issue a couple times, where we'll pass around a pre-built jar and different environments act differently by default. As it stands now, the responsibility lies on me to always remember when using docker java to pass `-Dfile.encoding=UTF8` (and it's probably good practice anyways). But again... whenever a bug like this pops up it takes me a couple hours to figure out what's going on and that's never fun. Should the docker java image have UTF8 by default?

See https://gist.github.com/devinrsmith/c2dddc652fbffe6404e9 for easy reproduction.

md5:
Previously, the image defaulted to ANSI_X3.4-1968

Closes #32

### DIFF
--- a/openjdk-6-jdk/Dockerfile
+++ b/openjdk-6-jdk/Dockerfile
@@ -7,6 +7,9 @@ FROM buildpack-deps:wheezy-scm
 
 RUN apt-get update && apt-get install -y unzip && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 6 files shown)


yosifkit:
LGTM

tianon:
SEXY
LGTM

PR #37, opened 2015-05-15
Fix for a single multi-line ENV statement

hilhorst:
My suggested changes for combining the separate `ENV` statements into one multi-line `ENV` statement, as proposed in https://github.com/docker-library/java/issues/35. Also updated `update.sh` and `generate-stackbrew-library.sh` to work with the new format (which breaks compatibility with the current format!).

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -25,7 +25,7 @@ for version in "${versions[@]}"; do
 	javaType="${javaVersion##*-}" # "jdk"
 	javaVersion="${javaVersion%-*}" # "6"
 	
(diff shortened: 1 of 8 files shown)


yosifkit:
That looks sane to me. Your thoughts, @tianon?

tianon:
I'm still a little cautious of this; the added script/regex complexity versus the gain of having three separate 0 byte layers be one single 0 byte layer seems like a dubious gain IMO.  It'd be really stellar if Docker could just see that there are three `ENV` instructions in a row and combine them into one layer automatically -- this is perhaps possible with the still-somewhat-recent builder rewrite.

hilhorst:
@tianon, like you once said, IANTM :wink: Feel free to close this pull request and the corresponding issue. I agree that the gain might not justify the added complexity. Special thanks to @md5 for taking the time to explain the build to me :thumbsup:

PR #38, opened 2015-05-16
Use Debian Stretch as base image for Java 8

md5:


yosifkit:
LGTM

tianon:
I'm torn on this one.  In theory, Stretch means less "unstable", but in practice, it really just means we have to wait for package migration to get new versions of OpenJDK. :confused:

md5:
@tianon No rush. I figured that opening the PR wouldn't hurt either way since this change will need to happen at some point. Feel free to let it ferment for a bit.

md5:
It probably makes sense to close this in favor of #40.

PR #40, opened 2015-05-21
Use Debian Jessie as base image for Java 8

md5:
This PR uses [jessie-backports](http://backports.debian.org/changes/jessie-backports.html) to get OpenJDK 8 under Debian Jessie.

I haven't yet made the necessary updates to `update.sh` since it's a bit tricky. The `docker run` statement that's used to check `apt-cache show` will need to conditionally add a Apt source list for `jessie-backports`. I could make it work, but it's going to be a hack.

### DIFF
--- a/openjdk-8-jdk/Dockerfile
+++ b/openjdk-8-jdk/Dockerfile
@@ -1,4 +1,4 @@
-FROM buildpack-deps:sid-scm
+FROM buildpack-deps:jessie-scm
 
 # A few problems with compiling Java from source:
 #  1. Oracle.  Licensing prevents us from redistributing the official JDK.
@@ -7,11 +7,13 @@ FROM buildpack-deps:sid-scm
 
 RUN apt-get update && apt-get install -y unzip && rm -rf /var/lib/apt/lists/*
 
+RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/jessie-backports.list
+
 # Default to UTF-8 file.encoding
 ENV LANG C.UTF-8
 
 ENV JAVA_VERSION 8u45
-ENV JAVA_DEBIAN_VERSION 8u45-b14-3
+ENV JAVA_DEBIAN_VERSION 8u45-b14-2~bpo8+2
 
 # see https://bugs.debian.org/775775
 # and https://github.com/docker-library/java/issues/19#issuecomment-70546872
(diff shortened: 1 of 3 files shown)


tianon:
https://github.com/docker-library/official-images/pull/762 might help :wink:

md5:
:metal:

md5:
@tianon Any chance you're thinking of doubling the number of tags on `buildpack-deps` to include `buildpack-deps:jessie-backports-curl` and `buildpack-deps:jessie-backports-scm`?

tianon:
Kind of wanting to avoid going _that_ crazy. :cry:

md5:
Then I'm thinking that `debian:jessie-backports` won't be much help here, unfortunately.

tianon:
I meant for "update.sh" -- it could assume that if
"debian:<suite>-backports" exists, it should scrape package version numbers
off that image instead of "buildpack-deps:<suite>".

tianon:
or "if grep -q '-backports' Dockerfile"

md5:
@tianon Good idea. That seems preferable to bloating the number of `buildpack-deps` tags.

md5:
Pushed 6e8a2bf266d877c00aae29c879165be7f76322b1 to update the `update.sh` script.

tianon:
LGTM

yosifkit:
LGTM

Godin:
While switching from `java:openjdk-8u45-jdk` to `java:8` we've noticed that this move to Jessie has a negative impact on curl+HTTPS ( https://github.com/SonarSource/docker-sonarqube/pull/18#issuecomment-194045499 ) due to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=812708 . Wondering if someone else faced the same problem? And what will be the best way to deal with this?

md5:
@Godin After reading through Debian issue 812708, it looks like the issue is that a number of 1024 bit certificates were removed from the `ca-certificates` package, including `Thawte_Premium_Server_CA.crt` and `GTE_CyberTrust_Global_Root.crt`.

It seems that these certificates were used to sign a number of other CA certificates that are still part of `ca-certificates`, so the child certificates should be trusted according the the specs. However, due to a couple of bugs in OpenSSL affecting versions older than 1.0.2, certificate chains using the intermediate CAs as the trust root can't be verified.

Since Debian `jessie` has version 1.0.1 of `openssl`, the combination of the CA removal and the OpenSSL bugs means that many certificate chains in the wild can no longer be verified.

The fix is to add the two removed certificates back. You should be able to get them from an older version of the `ca-certificates` package.

More information about the 1024 bit CA removal can be found here: https://blog.mozilla.org/security/2015/01/28/phase-2-phasing-out-certificates-with-1024-bit-rsa-keys/

PR #55, opened 2015-11-17
Add OpenJDK 9

tianon:


tianon:
Hahahahaha, `buildpack-deps:experimental-curl` doesn't exist.  Doh.

tianon:
Oh hah, we don't handle the case where `fullVersion` ends up being the raw number, since up to now we've only had `XuY` versions. :+1:

tianon:
Fix was pretty reasonable. :+1:

md5:
:metal:

tianon:
For reference: http://openjdk.java.net/projects/jdk9/ (has the timeline for release)

yosifkit:
LGTM

PR #61, opened 2016-01-05
Add explicit JAVA_HOME environment variable

### LINKED ISSUE #27: environment JAVA_HOME missing

tianon:
Closes #28 (commit carried/rebased here)
Fixes #27

For reference, the intended use-case of the "docker-java-home" script this installs is things like https://github.com/docker-library/tomcat/pull/19/files#diff-03f3e74233d756c542dc2d234fe3c43bR42, where we are `FROM` the JRE image, but install the JDK temporarily in order to compile something and then remove it again for size.

### DIFF
--- a/openjdk-8-jre/Dockerfile
+++ b/openjdk-8-jre/Dockerfile
@@ -12,6 +12,8 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
 # Default to UTF-8 file.encoding
 ENV LANG C.UTF-8
 
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+
 ENV JAVA_VERSION 8u66
 ENV JAVA_DEBIAN_VERSION 8u66-b17-1~bpo8+1
 
--- a/openjdk-8-jre/Dockerfile
+++ b/openjdk-8-jre/Dockerfile
@@ -18,6 +18,16 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
 # Default to UTF-8 file.encoding
 ENV LANG C.UTF-8
 
+# add a simple script that can auto-detect the appropriate JAVA_HOME value
+# based on whether the JDK or only the JRE is installed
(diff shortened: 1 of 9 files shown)


openjdk-9-jdk/Dockerfile
```diff
@@ -1,4 +1,10 @@
-FROM buildpack-deps:sid-curl
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "update.sh"
+#
+# PLEASE DO NOT EDIT IT DIRECTLY.
+#
+
+FROM buildpack-deps:sid-scm
```
tianon: Managed to catch an amusing oversight here, too. :smile:

openjdk-8-jre/Dockerfile
```diff
@@ -24,14 +42,11 @@ RUN set -x \
...
-	&& rm -rf /var/lib/apt/lists/*
+	&& rm -rf /var/lib/apt/lists/* \
+	&& [ "$JAVA_HOME" = "$(docker-java-home)" ]
 
 # see CA_CERTIFICATES_JAVA_VERSION notes above
 RUN /var/lib/dpkg/info/ca-certificates-java.postinst configure
 
-# see https://bugs.debian.org/793210
-# and https://github.com/docker-library/java/issues/46#issuecomment-119026586
-RUN apt-get update && apt-get install -y --no-install-recommends libfontconfig1 && rm -rf /var/lib/apt/lists/*
```
tianon: This bug is officially fixed, and the fix is included in the `openjdk-8` packages in `jessie-backports`. :+1:

md5: :metal:

tianon:
Yay, finally :green_apple:.

yosifkit:
LGTM

PR #62, opened 2016-01-11
add java:openjdk-8-jre-alpine

ncopa:
Not really sure how we want to do this.

I would also like to have a test that the ca-certificates actually works. (I think they currently don't work in Alpine)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
 
 before_script:
   - env | sort
-  - cd "${VERSION}${VARIANT:+-$VARIANT}"
+  - cd "${VERSION}${VARIANT:+/$VARIANT}"
   - image="java:${VERSION}${VARIANT:+-$VARIANT}"
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ services: docker
 env:
   - VERSION=openjdk-9-jre
   - VERSION=openjdk-9-jdk
+  - VERSION=openjdk-8-jre VARIANT=alpine
+  - VERSION=openjdk-8-jdk VARIANT=alpine
   - VERSION=openjdk-8-jre
   - VERSION=openjdk-8-jdk
   - VERSION=openjdk-7-jre
@@ -16,8 +18,8 @@ install:
 
 before_script:
   - env | sort
-  - cd "$VERSION"
-  - image="java:$VERSION"
+  - cd "${VERSION}${VARIANT:+-$VARIANT}"
+  - image="java:${VERSION}${VARIANT:+-$VARIANT}"
 
 script:
   - docker build -t "$image" .
--- a/openjdk-8-jdk/alpine/Dockerfile
+++ b/openjdk-8-jdk/alpine/Dockerfile
@@ -1,0 +1,29 @@
+FROM alpine:3.3
+
+# A few problems with compiling Java from source:
+#  1. Oracle.  Licensing prevents us from redistributing the official JDK.
+#  2. Compiling OpenJDK also requires the JDK to be installed, and it gets
+#       really hairy.
+
+# Default to UTF-8 file.encoding
+ENV LANG C.UTF-8
+
+# add a simple script that can auto-detect the appropriate JAVA_HOME value
+# based on whether the JDK or only the JRE is installed
+RUN { \
+		echo '#!/bin/sh -e'; \
+		echo; \
+		echo 'dirname "$(dirname "$(readlink -f "$(which javac || which java)")")"'; \
+	} > /usr/local/bin/docker-java-home \
+	&& chmod +x /usr/local/bin/docker-java-home
+
+ENV JAVA_HOME /usr/lib/jvm/java-1.8-openjdk
+
+ENV JAVA_VERSION 8u72
+ENV PATH $PATH:/usr/lib/jvm/java-1.8-openjdk/bin
+
+RUN set -x \
+	&& apk add --no-cache \
+		"openjdk8>=${JAVA_VERSION/u/.}" \
+		ca-certificates \
+	&& [ "$JAVA_HOME" = "$(docker-java-home)" ]
--- a/openjdk-8-jre/alpine/Dockerfile
+++ b/openjdk-8-jre/alpine/Dockerfile
@@ -1,0 +1,28 @@
+FROM alpine:3.3
+
+# A few problems with compiling Java from source:
+#  1. Oracle.  Licensing prevents us from redistributing the official JDK.
+#  2. Compiling OpenJDK also requires the JDK to be installed, and it gets
+#       really hairy.
+
+# Default to UTF-8 file.encoding
+ENV LANG C.UTF-8
+
+# add a simple script that can auto-detect the appropriate JAVA_HOME value
+# based on whether the JDK or only the JRE is installed
+RUN { \
+		echo '#!/bin/sh -e'; \
+		echo; \
+		echo 'dirname "$(dirname "$(readlink -f "$(which javac || which java)")")"'; \
+	} > /usr/local/bin/docker-java-home \
+	&& chmod +x /usr/local/bin/docker-java-home
+
+ENV JAVA_VERSION 8u72
+ENV JAVA_HOME /usr/lib/jvm/java-1.8-openjdk/jre
+
+RUN set -x \
+	&& apk add --no-cache \
+		"openjdk8>=${JAVA_VERSION/u/.}" \
+		ca-certificates \
+	&& [ "$JAVA_HOME" = "$(docker-java-home)" ]
+


openjdk-8-jdk-alpine/JavaHttpsTest.java
```diff
@@ -0,0 +1,27 @@
+/* JavaHttpsTest.java */
```
yosifkit: What is this file? It doesn't seem to be used anywhere.

yosifkit:
Useful diff:

```diff
diff --git a/jdk/Dockerfile b/alpine/Dockerfile
index f464a08..c5185b5 100644
--- a/Dockerfile
+++ b/Dockerfile.1
@@ -1,52 +1,28 @@
-#
-# NOTE: THIS DOCKERFILE IS GENERATED VIA "update.sh"
-#
-# PLEASE DO NOT EDIT IT DIRECTLY.
-#
-
-FROM buildpack-deps:jessie-scm
+FROM alpine:3.3
 
 # A few problems with compiling Java from source:
 #  1. Oracle.  Licensing prevents us from redistributing the official JDK.
 #  2. Compiling OpenJDK also requires the JDK to be installed, and it gets
 #       really hairy.
 
-RUN apt-get update && apt-get install -y unzip && rm -rf /var/lib/apt/lists/*
-
-RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/jessie-backports.list
-
 # Default to UTF-8 file.encoding
 ENV LANG C.UTF-8
 
 # add a simple script that can auto-detect the appropriate JAVA_HOME value
 # based on whether the JDK or only the JRE is installed
 RUN { \
-		echo '#!/bin/bash'; \
-		echo 'set -e'; \
+		echo '#!/bin/sh -e'; \
 		echo; \
 		echo 'dirname "$(dirname "$(readlink -f "$(which javac || which java)")")"'; \
 	} > /usr/local/bin/docker-java-home \
 	&& chmod +x /usr/local/bin/docker-java-home
 
-ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
-
+ENV JAVA_HOME /usr/lib/jvm/java-1.8-openjdk
 ENV JAVA_VERSION 8u66
-ENV JAVA_DEBIAN_VERSION 8u66-b17-1~bpo8+1
-
-# see https://bugs.debian.org/775775
-# and https://github.com/docker-library/java/issues/19#issuecomment-70546872
-ENV CA_CERTIFICATES_JAVA_VERSION 20140324
+ENV ALPINE_JAVA_VERSION 8.66.17-r2
+ENV PATH $PATH:/usr/lib/jvm/java-1.8-openjdk/bin
 
 RUN set -x \
-	&& apt-get update \
-	&& apt-get install -y \
-		openjdk-8-jdk="$JAVA_DEBIAN_VERSION" \
-		ca-certificates-java="$CA_CERTIFICATES_JAVA_VERSION" \
-	&& rm -rf /var/lib/apt/lists/* \
+	&& apk add --no-cache openjdk8=$ALPINE_JAVA_VERSION \
 	&& [ "$JAVA_HOME" = "$(docker-java-home)" ]
 
-# see CA_CERTIFICATES_JAVA_VERSION notes above
-RUN /var/lib/dpkg/info/ca-certificates-java.postinst configure
-
-# If you're reading this and have any feedback on how this image could be
-#   improved, please open an issue or a pull request so we can discuss it!
```

openjdk-8-jdk-alpine/JavaHttpsTest.java
ncopa: whoops. I used it to test that the cacerts handling actually works. Was not intended to be included.

ncopa:
rebased without the https test code

ncopa:
I brought it a bit closer to debians version:
````diff
--- openjdk-8-jdk/Dockerfile	2016-01-11 16:23:35.114547617 +0100
+++ openjdk-8-jdk-alpine/Dockerfile	2016-01-16 23:25:52.140392621 +0100
@@ -1,52 +1,29 @@
-#
-# NOTE: THIS DOCKERFILE IS GENERATED VIA "update.sh"
-#
-# PLEASE DO NOT EDIT IT DIRECTLY.
-#
-
-FROM buildpack-deps:jessie-scm
+FROM alpine:3.3
 
 # A few problems with compiling Java from source:
 #  1. Oracle.  Licensing prevents us from redistributing the official JDK.
 #  2. Compiling OpenJDK also requires the JDK to be installed, and it gets
 #       really hairy.
 
-RUN apt-get update && apt-get install -y unzip && rm -rf /var/lib/apt/lists/*
-
-RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/jessie-backports.list
-
 # Default to UTF-8 file.encoding
 ENV LANG C.UTF-8
 
 # add a simple script that can auto-detect the appropriate JAVA_HOME value
 # based on whether the JDK or only the JRE is installed
 RUN { \
-		echo '#!/bin/bash'; \
-		echo 'set -e'; \
+		echo '#!/bin/sh -e'; \
 		echo; \
 		echo 'dirname "$(dirname "$(readlink -f "$(which javac || which java)")")"'; \
 	} > /usr/local/bin/docker-java-home \
 	&& chmod +x /usr/local/bin/docker-java-home
 
-ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+ENV JAVA_HOME /usr/lib/jvm/java-1.8-openjdk
 
 ENV JAVA_VERSION 8u66
-ENV JAVA_DEBIAN_VERSION 8u66-b17-1~bpo8+1
-
-# see https://bugs.debian.org/775775
-# and https://github.com/docker-library/java/issues/19#issuecomment-70546872
-ENV CA_CERTIFICATES_JAVA_VERSION 20140324
+ENV JAVA_ALPINE_VERSION 8.66.17-r2
+ENV PATH $PATH:/usr/lib/jvm/java-1.8-openjdk/bin
 
 RUN set -x \
-	&& apt-get update \
-	&& apt-get install -y \
-		openjdk-8-jdk="$JAVA_DEBIAN_VERSION" \
-		ca-certificates-java="$CA_CERTIFICATES_JAVA_VERSION" \
-	&& rm -rf /var/lib/apt/lists/* \
+	&& apk add --no-cache openjdk8=$JAVA_ALPINE_VERSION \
 	&& [ "$JAVA_HOME" = "$(docker-java-home)" ]
 
-# see CA_CERTIFICATES_JAVA_VERSION notes above
-RUN /var/lib/dpkg/info/ca-certificates-java.postinst configure
-
-# If you're reading this and have any feedback on how this image could be
-#   improved, please open an issue or a pull request so we can discuss it!
````

**Comment by brunoborges** [comment]
The image glassfish/openjdk is based on Alpine and it works fine.

**Comment by tianon** [comment]
This makes me wonder if it's finally time to figure out compiling OpenJDK from source so that we can make sure that we keep parity between variants.

**Comment by ncopa** [comment]
> This makes me wonder if it's finally time to figure out compiling OpenJDK from source so that we can make sure that we keep parity between variants.

I doubt we want go there. To build openjdk you need a working jdk so you need bootstrap. gcc-java can be used for that to build openjdk7 but you need openjdk7 to build openjdk8. It is non-trivial.

**Comment by tianon** [comment]
Right, I'm aware it's non-trivial, but I think it's the only way we're going to get any degree of relative consistency between variants.

Requiring the language in use for building is relatively common (ref Go, Ruby, etc), and given that packages are available (but can't be expected to be consistent), I think this might be our only real viable option here.

**Comment by brunoborges** [comment]
@ncopa why you say OpenJDK 8 requires OpenJDK 7 to be built? The instructions are clear: 

http://hg.openjdk.java.net/jdk8/jdk8/raw-file/tip/README-builds.html

**Comment by tianon** [comment]
It does appear that way from the document you've linked:
```html
Install a
<a name="bootjdk">Bootstrap JDK</a>.
All OpenJDK builds require access to a previously released
JDK called the <i>bootstrap JDK</i> or <i>boot JDK.</i>
The general rule is that the bootstrap JDK
must be an instance of the previous major
release of the JDK. In addition, there may be
a requirement to use a release at or beyond a
particular update level.
<br>&nbsp;<br>

<b><i>Building JDK 8 requires use of a version
of JDK 7 that is at Update 7 or newer. JDK 8
developers should not use JDK 8 as the boot
JDK, to ensure that JDK 8 dependencies are
not introduced into the parts of the system
that are built with JDK 7.</i></b>

<br>&nbsp;<br>
The JDK 7 binaries can be downloaded from Oracle's 
<a href="http://www.oracle.com/technetwork/java/javase/downloads/index.html"
   target="_blank">JDK 7 download site</a>.
For build performance reasons
is very important that this bootstrap JDK be made available 
on the local disk of the machine doing the build.
You should add its <code>bin</code> directory
to the <code>PATH</code> environment variable.
If <code>configure</code> has any issues finding this JDK, you may
need to use the <code>configure</code> option
<code>--with-boot-jdk</code>.
```

(Although I would imagine that building JDK 8 with JDK 8 as bootstrap probably works too.)

tianon:
> JDK 8 developers should not use JDK 8 as the boot JDK, to ensure that JDK 8 dependencies are not introduced into the parts of the system that are built with JDK 7.

:confused:

brunoborges:
Yeah, sorry about missing that part.
Here is another document that might be useful: 
- https://java.net/projects/adoptopenjdk/pages/Build

ncopa:
@brunoborges because I worked with the people who built gcc-java, openjdk7 and openjdk8 with alpine. Building openjdk from sources is not fun.

ncopa:
> Right, I'm aware it's non-trivial, but I think it's the only way we're going to get any degree of relative consistency between variants.

I may have some influence on how the alpine openjdk package is built for alpine, in case that helps.

ncopa:
> (Although I would imagine that building JDK 8 with JDK 8 as bootstrap probably works too.)
yes , openjdk8 can build openjdk8.

@tianon if you still insist on re-building the openjdk from source then we could use the distro package as bootstrap jdk. Should simplify a bit. I'll follow if you do this with the debian based openjdk images.

That said, I don't think its worth the effort.

ncopa:
I rebased to patch against current master and did the following changes:

- fix directory structure
- fix travis to handle the alpine flavors
- update to 8u72
- add explicit version dependency to make sure it pulls in new enough openjdk

tianon:
Thanks for sticking with it @ncopa -- I'm satisfied with this for now, and we can put "compiling from source" on the backburner.  I'm a tiny bit concerned that the `>=` there might pull in a newer version some time in the future (which we'll then happily tag as `8u72`), but that shouldn't be too awful to mitigate either.

We'll definitely need some minor changes to `generate-stackbrew-library.sh`, and might need to think about what to do with `update.sh`, but I'm happy to take on those if you'd like.

LGTM cc @yosifkit for any further comments before merge

(I'd also note that the CI failure is thanks to https://bugs.debian.org/816440 -- this PR's bits succeeded)

ncopa:
The `>=` is not perfect but at least it make sure that you don't get old and potensial vulnerable. A small step forward.

@tianon would be great if you took the `generate-stackbrew-library.sh` and `update.sh`. thanks!

yosifkit:
LGTM

PR #68, opened 2016-02-25
clean out /var/cache/apt/archives before committing

cjac:
if you're going to remove the apt cache, you might as well clean out /var/cache/apt/archives as well.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -77,7 +77,7 @@ for version in "${versions[@]}"; do
 		#  2. Compiling OpenJDK also requires the JDK to be installed, and it gets
 		#       really hairy.
 
-		RUN apt-get update && apt-get install -y unzip && rm -rf /var/lib/apt/lists/*
+		RUN apt-get update && apt-get install -y unzip && apt-get clean && rm -rf /var/lib/apt/lists/*
 	EOD
 
 	if [ "$addSuite" ]; then
@@ -136,6 +136,7 @@ EOD
 EOD
 	fi
 	cat >> "$version/Dockerfile" <<EOD
+	&& apt-get clean \\
 	&& rm -rf /var/lib/apt/lists/* \\
 	&& [ "\$JAVA_HOME" = "\$(docker-java-home)" ]
 EOD


tianon:
This _should_ already be baked into our base image (`debian`; https://github.com/docker/docker/blob/6fa5576e308d4c99e9a818f924483536a773afba/contrib/mkimage/debootstrap#L82-L105) -- is that code not working?

cjac:
I'm not certain.  I hadn't tested it.  I was just reading through the code and saw that you're handling package index caches here.  I figured this would be the same way you'd handle packages.  I'll let you know once I perform the full build.  I'm trying to find a way to reduce the size of the jenkins docker image, since we host quite a few at linuxfoundation.org, and I am standing up some test environments.

jenkins                     latest              a61d8afe45ab        8 days ago          708.6 MB
local/jenkins/zephyr        latest              efc0dcebe6f3        19 hours ago        1.001 GB

I think I'm going to take your code from the above debootstrap link and drop it in to my local/jenkins/zephyr Dockerfile.  Thank you!

tianon:
Yeah, I can't get this to remove any space from the image I've got locally: (since we've already pre-configured `apt` to do this for us in the base image itself)
```console
$ docker run -it --rm --name test java bash
root@982941f24c51:/# apt-get clean
root@982941f24c51:/# 
$ # second window
$ docker diff test
$ 
```

cjac:
I am not very familiar with how Docker users Volumes yet, but perhaps there could be a shared

VOLUME        ["/var/cache/apt/archives"]

which catches all of the packages and is pruned from the image set before being pushed to the hub?

tianon:
If that were feasible, I'd love that, but unfortunately once we define that `VOLUME`, it cannot be undefined without squashing the image rootfs via `docker export`.  So, for your specific use case that might work, but cannot work in the general case. :disappointed:

cjac:
Okay.  Thanks for the quick feedback.  I'm happy to have this issue marked resolved.

tianon:
:+1:

PR #70, opened 2016-03-11
Expand included decompressors to explicitly include bzip2 and xz (on top of the existing zip)

### LINKED ISSUE #69: Include bzip2 package
Please consider including the bzip2 package. I understand it can be installed with apt-get, but it's very common to compress tarballs with bzip2 and it seems reasonable to support uncompressing those in the stock image.

tianon:
Fixes #69


yosifkit:
LGTM

yosifkit:
(build failure is a bug in Debian https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=816440)

PR #95, opened 2017-02-02
Update to 8u121, debian 8u121-b13-1~bpo8+1

antoineco:
openjdk `8u111-b14-2~bpo8+1` is no longer available in Debian Jessie, without this commit the `8-jre` and `8-jdk` flavours do not build.

---

I created a fork for the time being with automated builds, for those who need. See [comment](https://github.com/docker-library/openjdk/pull/95#issuecomment-277751517) below.

### DIFF
--- a/8-jdk/Dockerfile
+++ b/8-jdk/Dockerfile
@@ -34,12 +34,12 @@ RUN { \
 
 ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
 
-ENV JAVA_VERSION 8u111
-ENV JAVA_DEBIAN_VERSION 8u111-b14-2~bpo8+1
+ENV JAVA_VERSION 8u121
+ENV JAVA_DEBIAN_VERSION 8u121-b13-1~bpo8+1
 
 # see https://bugs.debian.org/775775
 # and https://github.com/docker-library/java/issues/19#issuecomment-70546872
-ENV CA_CERTIFICATES_JAVA_VERSION 20140324
+ENV CA_CERTIFICATES_JAVA_VERSION 20161107~bpo8+1
 
 RUN set -x \
 	&& apt-get update \
--- a/8-jre/Dockerfile
+++ b/8-jre/Dockerfile
@@ -34,12 +34,12 @@ RUN { \
 
 ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/jre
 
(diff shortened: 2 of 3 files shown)


leifcr:
@antoineco Updating CA_CERTIFICATES_JAVA_VERSION to 20161107 should fix the build.

antoineco:
@leifcr thanks for the hint, I'm looking into it.

antoineco:
I think I got it right. The PR is ready for review from my side 👌

cesar-oyarzun-m:
@antoineco do you have a work around for this, I have the same problem when I try to build my docker image.

antoineco:
I simply forked this repo and built the tag I needed locally. It's suboptimal if you build images with the `--pull` flag though.

If that helps I can set an automated build based on the proposed patch at 'amtoineco/openjdk'. Should I?

cesar-oyarzun-m:
I'm using circle ci to build the image just by doing this `sudo docker inspect -t exp-jcr-base -f docker/exp-jcr-base/Dockerfile  --quiet=false --rm=false `

cesar-oyarzun-m:
@antoineco I this already solved?, I don't know how I can use your repo? do you have instructions I can follow?. I just wants to install openjdk

antoineco:
@cesar-oyarzun-m if you just want to use openjdk as a base you don't need my repo at all. It is only useful if you have a dependency which relies on the packet version for compiling some java code using openjdk-jdk on top of the openjdk-jre image. The official tomcat image is the perfect example.

cesar-oyarzun-m:
Yeah I'm using the oficial tomcat image 7.0.72-jre8, I have the dependency for it , I'm just doing this `apt-get update && apt-get install -y --no-install-recommends openjdk-8-jdk maven` , I guess I need to copy your files and install java manually?, pretty new in here.

antoineco:
Hm, if you inherit from the tomcat image you also need a fixed tomcat image :)
The openjdk image was pushed to the hub some minutes ago. I'm sure tomcat will follow today.

yosifkit:
We have bumped the java version with https://github.com/docker-library/official-images/pull/2628 and are currently building dependent images, so tomcat should be rebuilt shortly.

cesar-oyarzun-m:
ok great @yosifkit so I just have to wait for the new Tomcat image and test it right?

PR #102, opened 2017-03-03
Update alpine (3.4 -> 3.5) & java (8u111 -> 8u121)

viartemev:


yosifkit:
We had tried moving to Alpine 3.5 when it came out, but reverted it with https://github.com/docker-library/openjdk/pull/92 since it could be a breaking change for many users.  That said, since this is the simplest way to get the Java 8 update via Alpine packages, I think we might be ok to bump the Alpine version since this will also be a bump on the Java version.

While I don't like breaking people but I am guessing the CVE's mentioned in the [Oracle patch advisory](http://www.oracle.com/technetwork/security-advisory/cpujan2017-2881727.html#AppendixJAVA) are more important and the users have a specific version jump to test the updated Alpine version.

cc @tianon

yosifkit:
If this is accepted, could you also update the `update.sh` script to ensure the right Alpine version when run by the bot? https://github.com/docker-library/openjdk/blob/7ed193579ce7dc6f4b69d31231568eea5adda600/update.sh#L20

viartemev:
@yosifkit done

tianon:
@viartemev in the future, please simply push new changes to the existing branch on your fork -- GitHub will automatically update this PR, which causes less churn in all our inboxes :+1:

PR #138, opened 2017-08-18
Alpine Openjdk 8u141

jpramondon:
Our Twistlock has recently reported a couple of security issues labelled as "high", which breaks our build. Most of the reported problems come from Openjdk version 8u131. Hence, this updated Dockerfile moves Openjdk to 8u141.
Additional apk repositories have been added in order to resolve all the required dependencies.

### DIFF
--- a/8-jdk/alpine/Dockerfile
+++ b/8-jdk/alpine/Dockerfile
@@ -19,6 +19,13 @@ FROM alpine:3.6
 # Default to UTF-8 file.encoding
 ENV LANG C.UTF-8
 
+# Add extra pak repositories that'll help resolving dependencies for version 8u141
+RUN echo "http://dl-1.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories; \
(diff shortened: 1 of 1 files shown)


tianon:
We've had a _lot_ of trouble in the past trying to mix released Alpine versions with packages from Edge, so I'd be pretty strongly opposed to this change.

If there are security vulnerabilities in the existing package, it shouldn't be hard to convince the Alpine package maintainers to do an update in Alpine 3.6 directly (they've done so in the past for exactly that reason). :+1:

joelbirchler:
I agree this should be fixed upstream by moving openjdk8-141 to Alpine v3.6 or an edge docker image could be built.

CVE: https://nvd.nist.gov/vuln/detail/CVE-2017-10102
Alpine package: https://pkgs.alpinelinux.org/package/edge/community/x86/openjdk8

tianon:
The OpenJDK 8 Alpine variants have since moved to Alpine 3.7, and thus 8u151 and this no longer applies. :+1:

PR #155, opened 2017-12-01
Update 8-jdk and 8-jre to alpine 3.7 and 8u151

wrporter:


yosifkit:
These changes will get reset by the jenkins job that runs `update.sh`

> ```dockerfile
> # NOTE: THIS DOCKERFILE IS GENERATED VIA "update.sh"
> #
> # PLEASE DO NOT EDIT IT DIRECTLY.
> ```
￼

wrporter:
@yosifkit Thanks for running the update script. Do these changes look good to merge?

PR #182, opened 2018-04-03
Use TLS 1.2 for GitHub downloads in Windows

tianon:
See https://github.com/docker-library/python/pull/228, https://github.com/docker-library/golang/pull/215, and https://githubengineering.com/crypto-removal-notice/.


tianon:
Updated to enable TLS 1.2 via registry on Nano Server instead, which builds successfully. :+1:

PR #160, opened 2018-01-17
Add OpenJDK 10 variants based on experimental

tianon:

PR #186, opened 2018-04-14
Add OpenJDK 11 variants based on experimental

risdenk:
Based on PR #160 for OpenJDK 11

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -18,6 +18,7 @@ declare -A suites=(
 	[8]='stretch'
 	[9]='sid'
 	[10]='sid'
+	[11]='sid'
 )
 declare -A alpineVersions=(
 	[7]='3.7'
@@ -27,7 +28,7 @@ declare -A alpineVersions=(
 
 declare -A addSuites=(
 	# there is no "buildpack-deps:experimental-xxx"
-	[10]='experimental'
+	#[11]='experimental'
 )
 
 declare -A buildpackDepsVariants=(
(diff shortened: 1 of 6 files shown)


risdenk:
Expecting that `jdk-11-jre*` image tests will fail since it requires `jdk-11-jdk` to compile it looks like. Same issue as in PR #160 for OpenJDK 10.

```
5.10s$ ~/official-images/test/run.sh "$image"
testing openjdk:11-jre
	'utc' [1/6]...passed
	'cve-2014--shellshock' [2/6]...passed
	'no-hard-coded-passwords' [3/6]...passed
	'override-cmd' [4/6]...passed
	'java-hello-world' [5/6]...Error response from daemon: manifest for openjdk:11-jdk not found
failed
	'java-uimanager-font' [6/6]...Error response from daemon: manifest for openjdk:11-jdk not found
failed
```

update.sh
```diff
@@ -28,6 +29,7 @@ declare -A alpineVersions=(
 declare -A addSuites=(
 	# there is no "buildpack-deps:experimental-xxx"
 	[10]='experimental'
+	[11]='experimental'
```
tianon: Looks like 10 and 11 are both in just `sid` now: https://packages.debian.org/source/sid/openjdk-11 (and https://packages.debian.org/source/sid/openjdk-10)

Can you update this to be just `#[11]='experimental'`? (as a reminder to us in the future)

tianon: (If you'd rather I take it from here, that's fine too. :+1:)

risdenk: Is that the same for [10]='experimental' since 10 is also just in sid?

Not sure which is easier for you. I don't mind updating :)

tianon: Yeah, sorry for not being more clear -- please remove both lines, and add just the comment to remind us what the expected format of this array is for the future (and don't forget to re-run the script to generate updated `Dockerfile`s). :+1:

risdenk:
Thanks @tianon!

PR #189, opened 2018-04-24
Disable assistive technologies in slim

carlossg:
Disable assistive technologies since slim installs headless JDK (without assistive technologies).

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -373,9 +373,15 @@ EOD
 		# for the "slim" variants,
 		#   - swap "buildpack-deps:SUITE-xxx" for "debian:SUITE-slim"
 		#   - swap "openjdk-N-(jre|jdk) for the -headless versions, where available (openjdk-8+ only for JDK variants)
+    #   - disable assistive technologies since headless JDK does not include them
+
+		assistiveDisable="# disable assistive technologies since headless JDK does not include them\\n \
+RUN sed -i 's/assistive_technologies=.*/assistive_technologies=/g' /etc/java-${javaVersion}-openjdk/accessibility.properties"
+
 		sed -r \
 			-e 's!^FROM buildpack-deps:([^-]+)(-.+)?!FROM debian:\1-slim!' \
 			-e 's!(openjdk-([0-9]+-jre|([89]\d*|\d\d+)-jdk))=!\1-headless=!g' \
+			-e "/update-alternatives --query java/a ${assistiveDisable}" \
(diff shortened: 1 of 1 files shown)


carlossg:
Brought to my attention by @MarkEWaite in https://github.com/jenkinsci/docker/pull/657

yosifkit:
Wouldn't anyone using openjdk headless from Debian (and derivatives like Ubuntu) have this same issue?  So if this is a problem in `accessibility.properties`, isn't this a bug in Debian's packaging of openjdk headless?

It looks like the cause is that `accessibility.properties` is provided by `openjdk-8-jre-headless` but is not overwritten by `openjdk-8-jdk-headless`, `openjdk-8-jre`, or `openjdk-8-jdk` (which all depend on `openjdk-8-jre-headless`).  I guess we could apply a temporary fix, but it should also be reported to the packagers so that it can be fixed in the right place.  After some limited poking into the Debian packaging, I think a quick fix could be to add stretch to [this line in `rules`](https://bazaar.launchpad.net/~openjdk/openjdk/openjdk8/annotate/head:/debian/rules#L1618), but that would probably break non-headless.

cc @tianon

MarkEWaite:
Yes, I believe that @yosifkit is correct that it is a packaging error in Debian.  A headless JRE is defined to not include assistive technologies and Debian packaging of a headless JRE is incorrectly enabling assistive technologies.

tianon:
Is this still an issue with the latest packages from upstream?

MarkEWaite:
I'm not sure what you precisely mean by "upstream", so I did a few checks:

```
$ docker run --rm -i -t openjdk:8-jdk-slim cat /etc/java-8-openjdk/accessibility.properties
#
# The following line specifies the assistive technology classes
# that should be loaded into the Java VM when the AWT is initailized.
# Specify multiple classes by separating them with commas.
# Note: the line below cannot end the file (there must be at
# a minimum a blank line following it).
#
assistive_technologies=org.GNOME.Accessibility.AtkWrapper
```

If "upstream" means the docker image "openjdk:8-jdk-slim", then the answer is "no".

If "upstream" means Debian slim, then the answer is "no" based on the state of the Debian bug report.

tianon:
"upstream" in this context is Debian's packages -- we consume those as-is, so if they're affected, we're affected.

See `openjdk:11-oracle` (https://github.com/docker-library/openjdk/pull/235) for the first foray into changing/fixing that relationship for 11+ (becoming the default variant in 12+).

PR #192, opened 2018-05-10
jdk-9 and jdk-10 with stable stretch

### LINKED ISSUE #179: Will there ever be a v9 image on a stable debian dist?
I'd like to use openjdk v9 but I have reservations about building my app on an "experimental" version of debian. Will there ever be a v9 on stretch? Curious as to the reasoning there.

Thank you for all you do to maintain these images, I'm happily using the v8 slim-stretch version and it works great.

arpitjindal97:
JDK9 has been [superseded](http://jdk.java.net/9/) and JDK10 is currently ready for [production](http://jdk.java.net/). 

So, I think it is not wrong to provide JDK10 with stable stretch images.

Fixes #179


tianon:
These packages don't exist in the stretch suite, which is why this can't work:

- https://packages.debian.org/sid/openjdk-9-jdk
- https://packages.debian.org/sid/openjdk-10-jdk

arpitjindal97:
Maybe, then we can use buster instead. They are available in buster suite. Atleast it is better then sid

- https://packages.debian.org/buster/openjdk-9-jdk
- https://packages.debian.org/buster/openjdk-10-jdk

PR #208, opened 2018-06-25
Moves jdk install to headless in 10/11 slim by fixing sed expression to properly match 10+

### LINKED ISSUE #207: JDK 10/11 slim do not use jdk-headless
https://github.com/docker-library/openjdk/blob/dd54ae37bc44d19ecb5be702d36d664fed2c68e4/10/jdk/slim/Dockerfile#L58

The current build script for publishing JDK 10 and 11 `slim` variants are not applying `-headless` to the script.  The JRE `slim` variants are not impacted:

https://github.com/docker-library/openjdk/blob/dd54ae37bc44d19ecb5be702d36d664fed2c68e4/10/jre/slim/Dockerfile#L58

yosifkit:
Fixes #207.

### DIFF
--- a/10/jdk/slim/Dockerfile
+++ b/10/jdk/slim/Dockerfile
@@ -55,7 +55,7 @@ RUN set -ex; \
 	\
 	apt-get update; \
 	apt-get install -y --no-install-recommends \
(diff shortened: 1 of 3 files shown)


yosifkit:
Build failures are versions missing in apt and will be addressed in a separate PR.

Issue #212, opened 2018-07-03
Use OpenJDK builds provided by jdk.java.net?


agudian:
With Oracle providing builds of OpenJDK (including updates) on http://jdk.java.net under the GPLv2 (with Classpath Exception), how about using those binaries the Docker images?

That would allow more timely updates and decouple the Java version updates from package and version cycles of the different Linux distros. That would be interesting especially for the Alpine variants that missed most of the last updates.

Or does anything speak against that?

(this is a bit more general than what's asked in #211)

agudian:
Of course it only makes sense on architectures that are supported by the builds, and keep using provided packages of the individual distributions otherwise.

And it looks like at least an additional Apline binary is already provided for the JDK 11 ea version, which is a good sign.

tianon:
That's definitely a no-go unless the provided packages match what the
distributions provide reasonably closely (and I don't believe that they
do); we try to keep the images for multiple architectures as similar as
possible to avoid unexpected issues due to architecture changes.

panga:
This will prevent issues like #217

tianon:
https://twitter.com/robilad/status/1032525402433101824?s=19 is a pretty compelling argument IMO, although losing so many arches still makes me a little sad -- what's the compelling story for those existing users? 😔

(ibmjava is probably a decent rec for IBM arches, but this would leave arm users out cold)

tianon:
Also worth pointing out this wouldn't provide jdk 7 or 8, but maybe that's not really all that important. 😅

tianon:
@Djelibeybi in thinking about this, I was considering adding Oracle Linux-based variants, and using those to provide the official upstream-provided JDK -- what are your thoughts around that?  That would help us navigate the backwards compatibility concerns, and make it clear this is something separate from what's been published so far.

Currently, our tags include the Debian release name (`stretch`, `jessie`, `alpine3.8`, etc), so for this I'd imagine adding something like `-oracle7`, etc. (or would we have to go all the way to `-oraclelinux7`?)

Djelibeybi:
Sorry @tianon, I meant to reply earlier but forgot. This week has been hectic.

We'd _prefer_ `oraclelinux7` but if making it shorter is required, please use `ol7` instead.

tianon:
It's not _required_ per se, but `oraclelinux7` is really a mouthful for users who want to use the upstream bits to have to type out explicitly. :smile:

[...]

PR #227, opened 2018-09-05
Add Windows bits for 11 and 12 from jdk.java.net (official upstream builds!)

tianon:
This is the first half of https://github.com/docker-library/openjdk/issues/212 -- Windows is easier, especially if we only do 11 and 12 (since those don't exist yet).

This also allows Windows to get a one-up on Debian and Alpine by supporting 12. :muscle:

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,6 +3,10 @@ image: Visual Studio 2017
 
 environment:
   matrix:
+    - version: 12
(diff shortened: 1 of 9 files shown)


tianon:
I wanted to do 10 as well, but the builds from http://jdk.java.net/10/ for Windows are `.tar.gz` files which are non-trivial for us to extract here. :sweat_smile:

Issue #243, opened 2018-10-12
JVM (OpenJDK 11) fails to start when JFR is enabled


laszlocsontos:
This is a bug in the current `openjdk-11-jdk` package Debian provides.
Related bug report: [JVM fails to start when JFR is enabled](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=910804)
------
**How to reproduce**

```
% docker run openjdk:11 \
  java -XX:StartFlightRecording=disk=true,dumponexit=true,filename=/tmp/test.jfr -version

Could not parse setting default
Error occurred during initialization of VM
Failure when starting JFR on_vm_start
```

**Analysis**

The problem is caused by missing JFC files from the OpenJDK installation.

Another JDK installation which I created form the official tarball has
the following files:

```
# ls -l /opt/jdk-11/lib/jfr/
total 56
-rw-r--r-- 1 668 668 28012 Aug 23 03:57 default.jfc
-rw-r--r-- 1 668 668 27963 Aug 23 03:57 profile.jfc
```

On the other hand, that one I'm having the problem with doesn't even have
the JFR folder

```
# ls -l /usr/lib/jvm/java-11-openjdk-amd64/lib/jfr
ls: cannot access '/usr/lib/jvm/java-11-openjdk-amd64/lib/jfr': No such file or directory
```

**Wordaround**

Build a custom container image and add the following `RUN` statement after `FROM openjdk:11`.

```
# Workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=910804
RUN mkdir -p /docker-java-home/lib/jfr && \
    wget https://hg.openjdk.java.net/jdk/jdk11/raw-file/76072a077ee1/src/jdk.jfr/share/conf/jfr/default.jfc \
      -O /docker-java-home/lib/jfr/default.jfc && \
    wget https://hg.openjdk.java.net/jdk/jdk11/raw-file/76072a077ee1/src/jdk.jfr/share/conf/jfr/profile.jfc \
      -O /docker-java-home/lib/jfr/profile.jfc
```

The default JFR profile will be found after that.

```
% docker run 637bba0e215b java -XX:StartFlightRecording=disk=true,dumponexit=true,filename=/tmp/test.jfr -version 
Started recording 1. No limit specified, using maxsize=250MB as default.

Use jcmd 1 JFR.dump name=1 to copy recording data to file.
openjdk version "11" 2018-09-25
[...]

PR #244, opened 2018-10-12
JVM (OpenJDK 11) fails to start when JFR is enabled

laszlocsontos:
Related issue: https://github.com/docker-library/openjdk/issues/243


yosifkit:
I would rather wait for the fix to come through from Debian's packages.

On a side note, the download appears to be a random commit on `hg.openjdk.java.net` but the confusion could be mitigated by using the hg tag `jdk-11+28` in the url instead.

Issue #241, opened 2018-10-10
Windows 1803 Images for JDK <11 


el-gee:
It would be wonderful to have prebuilt 1803 images for the older java versions (in my case specifically 8).  A lot of enterprise applications still only support java 8 but benefit immensely from the latest windows container versions.  

From my initial testing it seems as simple as a copy paste of the 1709 Dockerfile and switching the FROM line is sufficient.

mixomat:
Thank you @tianon for merging that PR. Is there another step required to publish the images to dockerhub?

tianon:
Yeah: https://github.com/docker-library/openjdk#see-a-change-merged-here-that-doesnt-show-up-on-docker-hub-yet :wink:

(I'm making a PR with the change now. :+1:)

PR #248, opened 2018-11-07
Add jdk8 variant for windowsservercore-1803

mixomat:


mixomat:
Fixes https://github.com/docker-library/openjdk/issues/241

PR #252, opened 2018-11-30
Create Dockerfile for nanoserver:sac2016 for openjdk 11

Pvlerick:
This is basically a copy of https://github.com/docker-library/openjdk/blob/master/10/jdk/windows/nanoserver-sac2016/Dockerfile updating versions numbers to get the correct jdk version.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,6 +7,8 @@ environment:
       variant: windowsservercore-ltsc2016
     - version: 11
       variant: windowsservercore-ltsc2016
+    - version: 11
+      variant: nanoserver-sac2016
     - version: 10
       variant: windowsservercore-ltsc2016
     - version: 10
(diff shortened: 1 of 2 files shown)


Pvlerick:
Got to update the CI files still.

tianon:
Honestly, I wasn't sure anyone was even still using `nanoserver:sac2016` given the availability of the semi-annual releases of it which are fundamentally incompatible to the point of not being something we can reasonably consume (which is why they weren't included/updated here).

I will say that for this to be something we can consider, it needs to be in line with https://github.com/docker-library/openjdk/pull/227 (not the old ojdkbuild-based variants).

Pvlerick:
Well, I use the sac2016 because of what you pointed out: later releases have been stripped of pretty much anything and are unusable for anything else than dotnet, basically.

I'm trying to stay away from windowsservercore when I can, but it looks like I won't have a choice...

PR #254, opened 2018-12-06
Remove EOL 10 variants

tianon:
Debian has removed openjdk-10 from the Debian archive (which causes those to fail), and Oracle no longer supports it (to my knowledge).


tianon:
FYI @carlossg @makuk66 @shalinmangar @gfinocchiaro

PR #269, opened 2019-01-15
Add support for Windows Server 1809 (LTSC 2019)

ayfieipatiosasmanidis:


tianon:
See also https://github.com/docker-library/golang/pull/257#issuecomment-449772799 -- we've decided to call this build `1809` (since `ltcs2019` and `1809` are currently the same for the container base and `1809` is the more specific designation of the two).  I'll take over so we can push this across all supported versions too; thanks! :+1:

PR #276, opened 2019-01-31
Update to Alpine 3.9

J0WI:


generate-stackbrew-library.sh
```diff
@@ -196,7 +196,7 @@ for javaVersion in "${versions[@]}"; do
 					}
 					fromTag = $2
 				}
-				$1 == "RUN" && $2 == "echo" && $4 == "http://deb.debian.org/debian" && $5 !~ /-backports$/ {
+				$1 == "RUN" && $2 == "echo" && $4 == "https://deb.debian.org/debian" && $5 !~ /-backports$/ {
```
tianon: This change definitely isn't right. :confused:

J0WI: What's wrong with it?

tianon: It's matching a line that it expects to find inside the Dockerfile, which will not have https.

J0WI: Ok, I see. Fixed.

J0WI:
I'll rebase in a minute

Issue #272, opened 2019-01-17
Update to version X on base os Y


yosifkit:
Since we consume OpenJDK mainly from upstream or trusted/TCK tested binaries, here's a list of what can be updated and how.

Most version updates are handled automatically by `update.sh` run via Jenkins (see README for last build status). OS distribution bumps are handled via an update to the corresponding template files in the root. We will be unlikely to move to Debian testing, Debian Sid or Alpine Edge, since all of these are "unstable" distributions.

- openjdk 7
  - no more OpenJDK 7 images (see https://github.com/docker-library/openjdk/issues/320#issuecomment-494417274 -- "7 is still alive but only receives security updates" and "compatibility between 7 and 8 seems to be good enough in most cases") https://github.com/docker-library/openjdk/pull/322
- openjdk 8
  - JDK/JRE
    - Openjdk 8 and 11 come from https://adoptopenjdk.net/upstream.html which does publish a JRE variant
  - Alpine
    - not officially supported by the OpenJDK project https://github.com/docker-library/openjdk/pull/322, https://github.com/docker-library/openjdk/issues/211#issuecomment-426708580, https://github.com/docker-library/openjdk/pull/235#issuecomment-424599754
  - Debian & Windows
    - uses latest builds from https://adoptopenjdk.net/upstream.html
- openjdk 11
  - JDK/JRE
    - Openjdk 8 and 11 come from https://adoptopenjdk.net/upstream.html which does publish a JRE variant https://github.com/docker-library/openjdk/commit/572682ed1cb951ab8d978e8015e4a910093db920
  - Alpine
    - not officially supported by the OpenJDK project https://github.com/docker-library/openjdk/issues/211#issuecomment-426708580, https://github.com/docker-library/openjdk/issues/212#issuecomment-424453577, https://github.com/docker-library/openjdk/pull/235#issuecomment-424599754
  - Debian & Windows
    - uses latest builds from https://adoptopenjdk.net/upstream.html
- openjdk 12
  - has reached end of life (https://en.wikipedia.org/wiki/Java_version_history) https://jdk.java.net/12/
- openjdk 13
  - has reached end of life (https://en.wikipedia.org/wiki/Java_version_history) https://jdk.java.net/13/
- openjdk 14
  - has reached end of life (https://en.wikipedia.org/wiki/Java_version_history) https://jdk.java.net/14/ 
- openjdk 15
[...]

PR #235, opened 2018-09-25
Add initial all-Oracle variants (oraclelinux + jdk.java.net tarballs)

tianon:
Closes #212
Closes https://github.com/docker-library/openjdk/issues/236

FYI @Djelibeybi (any and all comments, thoughts, review, etc would be super appreciated :heart:)

Issue #211, opened 2018-07-02
Build Java 11 Alpine


ST-DDT:
On the java page there is a alpine compatible build for Java 11.

http://jdk.java.net/11/
https://download.java.net/java/early_access/alpine/19/binaries/openjdk-11-ea+19_linux-x64-musl_bin.tar.gz

Could you please build a docker image for that, so I could start testing Java 11 before Java 8 goes EOL this year?

Or is it required to appear in the official repository that was linked in some other issues https://pkgs.alpinelinux.org/packages?name=openjdk*&branch=edge ?

tianon:
Isn't `src.zip` useful to include for many use cases?  (Especially since this is already the full JDK, so not exactly "slim".)

For context, the current `openjdk:11-jdk` image contains `src.zip`:
```console
$ docker run --rm openjdk:11-jdk find /usr -name src.zip
/usr/lib/jvm/java-11-openjdk-amd64/src.zip
```

tianon:
I'm a little bit concerned that this will be a divergence from our previous Alpine images (and will mean we can only support amd64), but given that https://bugs.alpinelinux.org/issues/8089 is still open and unresolved 9 months and 2 releases later, I think moving forward with the upstream-provided binaires for Alpine probably makes sense.

tianon:
(It looks like we could even add the Early Access builds of OpenJDK 12 via this method, which would get the jump on Debian too.)

tianon:
This is related to https://github.com/docker-library/openjdk/issues/212 too 👍

nafeger:
The openjdk build for alpine was removed for jdk 11, there is only an EAP for jdk12. Should the default download for linux jdk work on a stock alpine image?

tianon:
Yeah, see https://github.com/docker-library/openjdk/issues/212#issuecomment-424453577 and https://github.com/docker-library/openjdk/pull/235#issuecomment-424599754:

> Yeah, the alpine port is not in a supported release, since it's not in the mainline code base, so it's only available as early access builds of OpenJDK Project Portola http://openjdk.java.net/projects/portola.  
> 
> It works, but at this stage it basically needs community contributors to help maintain it and get it into mainline - see https://twitter.com/MikaelVidstedt/status/984920426655526912 for details. Interested contributors should drop by portola-dev to discuss specifics.

nafeger:
[...]

PR #281, opened 2019-02-14
Drop 12-alpine (matching https://jdk.java.net/12/)

tianon:
It's a shame, but it appears OpenJDK has dropped Alpine builds from 12 (https://jdk.java.net/12/).

Refs https://github.com/docker-library/openjdk/issues/211#issuecomment-426708580, https://github.com/docker-library/openjdk/issues/212#issuecomment-424453577, https://github.com/docker-library/openjdk/pull/235#issuecomment-424599754, #272


tianon:
Oh, this was because 12 is now a proper release candidate! :+1: :metal:

tianon:
(So Portola must still be in need of help -- please see the references in my original comment if you want to help out to make Alpine a properly supported target in OpenJDK!)

PR #291, opened 2019-03-23
Add java 8 en java 11 images for nanoserver 1809

esrahofstede:
Adding java 8 and 11 images for windows nanoserver 1809
There are currently only Windows Server Core 2016 images available for java 8 and 11.
Since nanoserver is smaller and faster than Windows Server Core, these images bring a better performance.

The images are build multistage, so they are as small as possible.
The verify installer step checks if java and javac can be called from the command line and exists the build process if not found.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -19,6 +19,8 @@ environment:
       variant: windowsservercore-ltsc2016
     - version: 8
       variant: nanoserver-sac2016
+    - version: 8
+      variant: nanoserver-1809
 
 install:
   - ps: |
--- a/11/jdk/windows/nanoserver-1809/Dockerfile
+++ b/11/jdk/windows/nanoserver-1809/Dockerfile
@@ -1,0 +1,63 @@
+FROM mcr.microsoft.com/powershell:6.1.3-nanoserver-1809	as builder
+
+# $ProgressPreference: https://github.com/PowerShell/PowerShell/issues/2138#issuecomment-251261324
+SHELL ["pwsh", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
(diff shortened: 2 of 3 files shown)


esrahofstede:
Fixes #293

tianon:
See https://github.com/docker-library/official-images/issues/3383 for the discussion around multi-stage builds in the official images program.

tianon:
Closing given that this is not something we can reasonably support right now (given our downstream constraints).

If someone wants to maintain this externally with minimal effort, it should *theoretically* be possible to do `COPY --from=openjdk:11-windowsservercore ...` or `COPY --from=openjdk:11-nanoserver ...` (although that's unverified).

esrahofstede:
Since https://github.com/docker-library/official-images/pull/5929 is merged, can we use multi-stage builds now? The faq is also updated: https://github.com/docker-library/faq#multi-stage-builds

tianon:
Fair enough.

I did some experimentation with this, and I'd say it's now blocked on https://bugs.openjdk.java.net/browse/JDK-8218486 / https://bugs.openjdk.java.net/browse/JDK-8225425 (which I ran into immediately after getting an image with Java in it).

tianon:
I got it on OpenJDK 8, 11, 13, etc (version doesn't matter -- it seems OpenJDK doesn't actually support Nano Server):

```
Launching JShell execution engine threw: C:\openjdk-13\bin\net.dll: Can't find dependent libraries
```

tianon:
If anyone wants to play with where I left off, my changes are pushed to https://github.com/docker-library/openjdk/compare/master...infosiftr:nanoserver

tianon:
Closing this in favor of https://github.com/docker-library/openjdk/issues/293 for tracking the status of Nano Server support in the OpenJDK image instead.

PR #294, opened 2019-03-26
Force install nss with openjdk in alpine

### LINKED ISSUE #289: Missing libnss3.so after updating openjdk:8-jdk-alpine image
Hello,

I have a docker file that builds openjdk:8-jdk-alpine based images. When updating the image internally (from the docker file) using `apk update && apk upgrade ` , and running my (spring-boot) app, it throws `java.io.IOException: Error loading shared library libnss3.so: No such file or directory`. However; after removing the `apk update && apk upgrade ` and building the image again, the app runs smoothly. 

I have logged into my images in both cases, and made sure that the file wasn't there on the first case, and was there after removing `apk update && apk upgrade `. In the later case, it was located in `/usr/lib/`

Is this a bug, or can you please inform me of what I might be missing ?

luiszimmermann:
Adding nss to the alpine image

Closes #289

### DIFF
--- a/8/jdk/alpine/Dockerfile
+++ b/8/jdk/alpine/Dockerfile
@@ -31,13 +31,16 @@ RUN { \
 ENV JAVA_HOME /usr/lib/jvm/java-1.8-openjdk
 ENV PATH $PATH:/usr/lib/jvm/java-1.8-openjdk/jre/bin:/usr/lib/jvm/java-1.8-openjdk/bin
 
-ENV JAVA_VERSION 8u191
-ENV JAVA_ALPINE_VERSION 8.191.12-r0
+ENV JAVA_VERSION 8u201
+ENV JAVA_ALPINE_VERSION 8.201.08-r0
 
 RUN set -x \
 	&& apk add --no-cache \
 		openjdk8="$JAVA_ALPINE_VERSION" \
 	&& [ "$JAVA_HOME" = "$(docker-java-home)" ]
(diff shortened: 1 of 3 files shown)


matsewe:
The problem concerning the jessie-update repository which led to build failures for `VERSION=7` has been fixed; builds should pass now.

luiszimmermann:
Triggering Travis

matsewe:
I guess, another run of `update.sh` should fix this problem. Most recent version of opendjk-7 for `jessie` is, according to `update.sh`, `7u211-2.6.17-1~deb8u1`. This is not yet incorporated into the Dockerfile.

tianon:
I've updated this with the minimal change and cancelled every job on Travis (and AppVeyor) that isn't related to the actual fix, so once https://travis-ci.org/docker-library/openjdk/jobs/512206934 is green we're good to go (assuming @yosifkit approves of my approach).

tianon:
(Then @docker-library-bot will take care of testing and pushing the rest of the updates as appropriate.)

tianon:
Welp, since Travis is having a sad, I've built both of these successfully locally and run the `java-ca-certificates` test directly and they both now pass.

J0WI:
This has been fixed upstream in https://git.alpinelinux.org/aports/commit/?h=3.9-stable&id=87ea954c9b7e200efbb52b4395dfc160fa4f0b55

PR #305, opened 2019-04-18
Remove all 11-oracle variants (https://jdk.java.net/11/ EOL)

tianon:
Was thinking it might also make sense to remove the other `11` variants, but maybe those can stay until Debian stops supporting them?  Maybe move them from `sid` to `buster` once it goes stable?  Not sure what else/more to do around transition here. :thinking:


tianon:
To quote https://jdk.java.net/11/ for posterity:

> The JDK 11 Early Access Program has concluded.

PR #315, opened 2019-05-11
Default to a sensible encoding

ekoutanov:
Will work with files that contain unicode characters, closest to Debian's`C.UTF-8` on distros where the latter _tends_ to be omitted out of the box (Arch, Fedora, CentOS, RedHat, Oracle, etc.). 

Note: The default (if `LANG` is unspecified) is just `en_US` (well, actually depending on the installation locale), which is ISO 8859-1 — essentially ASCII with an 8th bit. Given Java's native support for unicode, leaving the default `en_US` is less than ideal.

### DIFF
--- a/12/jdk/oracle/Dockerfile
+++ b/12/jdk/oracle/Dockerfile
@@ -12,8 +12,7 @@ RUN set -eux; \
 	rm -rf /var/cache/yum
 
 # Default to UTF-8 file.encoding
-#ENV LANG C.UTF-8
-# TODO oraclelinux doesn't have C.UTF-8 by default??
+ENV LANG=en_US.UTF-8
 
 ENV JAVA_HOME /usr/java/openjdk-12
 ENV PATH $JAVA_HOME/bin:$PATH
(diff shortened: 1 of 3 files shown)


tianon:
There was some discussion around this over in https://github.com/docker-library/openjdk/issues/212#issuecomment-421130913 before we created these images -- @robilad do you think this makes sense while https://openjdk.java.net/jeps/8187041 is still in-progress?  Do you think it would make sense to link to https://openjdk.java.net/jeps/8187041 in a comment here (both as a reminder for us and as a pointer for users looking for further reading)?

ekoutanov:
The issue you refer to is unfortunately closed. Seems to have been a passing conversation that wasn't really addressed at the time.

With the current defaults, `javac` complains when compiling a file that has unicode characters, so clearly the defaults are insufficient.

ekoutanov:
So what are we doing about this? @tianon, noticed you reacted with a confused emoji; is there anything you need me to clarify about the pull request?

tianon:
Apologies, I lost track of this one.  I've just pushed a rebase and an extra commit to update the template and version 13; if the CI passes, I think we're good to go. :+1:

PR #133, opened 2017-07-21
Add slim variants

tianon:
Closes #131

cc @bjornbak :+1:

I think we should consider swapping the non-`slim` variants to not use `-headless` as well, but IMO that should happen as a separate discussion for this one (I only bring it up because having this is a precursor to being able to consider something like that, ala #130). :smile:

PR #322, opened 2019-05-22
Move 8 and 11 to consume from https://adoptopenjdk.net/upstream.html

### LINKED ISSUE #73: NullPointerException in Alpine JRE 8 Font
JVM Version: 1.8.0_77-internal-alpine-r0-b03

```
INFO [CaptchaImpl] Initializing com.liferay.portal.captcha.simplecaptcha.SimpleCaptchaImpl
Fontconfig warning: ignoring C.UTF-8: not a valid language tag
ERROR [CaptchaPortletAction] null
java.lang.NullPointerException
    at sun.awt.FontConfiguration.getVersion(FontConfiguration.java:1264)
    at sun.awt.FontConfiguration.readFontConfigFile(FontConfiguration.java:219)
    at sun.awt.FontConfiguration.init(FontConfiguration.java:107)
    at sun.awt.X11FontManager.createFontConfiguration(X11FontManager.java:774)
    at sun.font.SunFontManager$2.run(SunFontManager.java:431)
    at java.security.AccessController.doPrivileged(Native Method)
    at sun.font.SunFontManager.<init>(SunFontManager.java:376)
    at sun.awt.FcFontManager.<init>(FcFontManager.java:35)
    at sun.awt.X11FontManager.<init>(X11FontManager.java:57)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
    at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
    at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
    at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
    at java.lang.Class.newInstance(Class.java:442)
    at sun.font.FontManagerFactory$1.run(FontManagerFactory.java:83)
    at java.security.AccessController.doPrivileged(Native Method)
    at sun.font.FontManagerFactory.getInstance(FontManagerFactory.java:74)
    at java.awt.Font.getFont2D(Font.java:491)
    at java.awt.Font.access$000(Font.java:224)
    at java.awt.Font$FontAccessImpl.getFont2D(Font.java:228)
    at sun.font.FontUtilities.getFont2D(FontUtilities.java:180)
    at sun.font.StandardGlyphVector.initFontData(StandardGlyphVector.java:1126)
    at sun.font.StandardGlyphVector.init(StandardGlyphVector.java:1115)
    at sun.font.StandardGlyphVector.<init>(StandardGlyphVector.java:167)
    at java.awt.Font.createGlyphVector(Font.java:2545)
    at nl.captcha.text.renderer.DefaultWordRenderer.render(Unknown Source)
    at nl.captcha.Captcha$Builder.addText(Unknown Source)
    at com.liferay.portal.captcha.simplecaptcha.SimpleCaptchaImpl.getSimpleCaptcha(SimpleCaptchaImpl.java:243)
    at com.liferay.portal.captcha.simplecaptcha.SimpleCaptchaImpl.serveImage(SimpleCaptchaImpl.java:159)
    at com.liferay.portal.captcha.CaptchaImpl.serveImage(CaptchaImpl.java:100)
    at com.liferay.portal.kernel.captcha.CaptchaUtil.serveImage(CaptchaUtil.java:78)
    at com.liferay.portal.captcha.CaptchaPortletAction.serveResource(CaptchaPortletAction.java:42)
```

### LINKED ISSUE #217: jlink runtime image size (specifically libjvm.so) created from openjdk is huge. Expected it to be much smaller.
**Description** 

When building a Docker image for a service based on OpenJDK base image, I noticed that the runtime image created is `346Mb` (and specifically the file `libjvm.so` under /lib/security) in size as compared to OracleLinux base image which creates a runtime image of `49Mb`. What is the reason for this huge disparity in size?

I have tried building from different variants of 10 and 11 base images, but I still see the large runtime size.

**Configuration**

Docker -
```
Client:
 Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   9ee9f40
 Built:        Thu Apr 26 07:13:02 2018
 OS/Arch:      darwin/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.03.1-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.5
  Git commit:   9ee9f40
  Built:        Thu Apr 26 07:22:38 2018
  OS/Arch:      linux/amd64
  Experimental: true
```
macOS - Sierra (10.12.6)

Dockerfile -

```
FROM openjdk:10.0.2-slim

RUN jlink \
    --add-modules java.xml.bind,java.sql,java.naming,java.management,java.instrument,java.security.jgss \
    --verbose \
    --strip-debug \
    --compress 2 \
    --no-header-files \
    --no-man-pages \
    --output /opt/jre-minimal

CMD du -sh /opt/jre-minimal
```

**Actual Results**
```
docker run --name ojdk -it openjdk_test
346M	/opt/jre-minimal
```

Digging a little deeper, it appears the primary reason for this bloat is `libjvm.so`.
```
docker run --rm --name openjdk_image_size -it openjdk_test ls -la /opt/jre-minimal/lib/server
total 308028
drwxr-xr-x 2 root root      4096 Jul 23 17:58 .
drwxr-xr-x 5 root root      4096 Jul 23 17:58 ..
-rw-r--r-- 1 root root      1322 Jul 23 17:58 Xusage.txt
-rw-r--r-- 1 root root     24992 Jul 23 17:58 libjsig.so
-rw-r--r-- 1 root root 315379056 Jul 23 17:58 libjvm.so
```

**Expected Results**
Expect a much smaller runtime image. For example if I use [Oracle Linux](https://hub.docker.com/_/oraclelinux/) as base image,

Dockerfile -
```
FROM oraclelinux:7-slim

RUN yum -y install wget tar gzip && yum clean all \
    && wget -q https://download.java.net/java/GA/jdk10/10.0.2/19aef61b38124481863b1413dce1855f/13/openjdk-10.0.2_linux-x64_bin.tar.gz \
        -P /opt/jdk \
    && tar -xzf /opt/jdk/openjdk-10.0.2_linux-x64_bin.tar.gz -C /opt/jdk \
    && rm /opt/jdk/openjdk-10.0.2_linux-x64_bin.tar.gz

ENV PATH=${PATH}:/opt/jdk/jdk-10.0.2/bin

RUN jlink \
    --add-modules java.xml.bind,java.sql,java.naming,java.management,java.instrument,java.security.jgss \
    --verbose \
    --strip-debug \
    --compress 2 \
    --no-header-files \
    --no-man-pages \
    --output /opt/jre-minimal

CMD du -sh /opt/jre-minimal
```

I get a `49Mb` image size, which is expected.
```
docker run --name oracle_linux_jdk -it oraclelinux_jdk_test
49M	/opt/jre-minimal
```

### LINKED ISSUE #278: Getting different architectures in slim images
Slim images give the wrong architecture, see below:

```sh
pi@raspberrypi:~ $ docker run -it arm32v7/openjdk:8-jre-slim-stretch /bin/sh
# dpkg --print-architecture
armel

pi@raspberrypi:~ $ docker run -it arm32v7/openjdk:8-jre-stretch /bin/sh
# dpkg --print-architecture
armhf

pi@raspberrypi:~ $ docker run -it arm32v7/openjdk:8-jdk-slim /bin/sh
# dpkg --print-architecture
armel

pi@raspberrypi:~ $ dpkg --print-architecture
armhf
```

### LINKED ISSUE #303: openjdk:8-jre-alpine contains CVE-2018-14498
Description:
get_8bit_row in rdbmp.c in libjpeg-turbo through 1.5.90 and MozJPEG through 3.3.1 allows attackers to cause a denial of service (heap-based buffer over-read and application crash) via a crafted 8-bit BMP in which one or more of the color indices is out of range for the number of palette entries.

CVSS v3.0 Severity and Metrics:
Base Score: 6.5 MEDIUM 
Vector: AV:N/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H (V3 legend) 
Impact Score: 3.6 
Exploitability Score: 2.8

NVD Reference:
https://nvd.nist.gov/vuln/detail/CVE-2018-14498

I see that there is a patch for this vulnerability. Will it be applied to the openjdk:8-jre-alpine docker image in dockerhub?

### LINKED ISSUE #309: Travis Build failing with arm32v7/openjdk:11
Travis Build fails with this error:

```
Sending build context to Docker daemon  452.2MB
Step 1/12 : FROM arm32v7/openjdk:11
11: Pulling from arm32v7/openjdk
no matching manifest for unknown in the manifest list entries
```

You can find more log here: https://api.travis-ci.com/v3/job/197770047/log.txt

I pulled the image on my raspberry pi without any issues. Can someone explain to me why this is happening?

### LINKED ISSUE #320: OpenJDK 8u212 and 11.0.3 image quality concerns
I'm reading

https://mail.openjdk.java.net/pipermail/jdk8u-dev/2019-May/009330.html

entitled "Mystery meat OpenJDK builds strike again", in which it is noted that the `openjdk:8` and `openjdk:11` versions were observed to be not release versions but instead pre-release versions, originating from Debian and not reflecting the actual final releases of those versions.

tianon:
Closes #320

> These binaries are built by Red Hat on behalf of the OpenJDK jdk8u and jdk11u projects. The binaries are created from the unmodified source code at OpenJDK.

So to put that another way and be explicitly clear, the builds from https://adoptopenjdk.net/upstream.html are *not* AdoptOpenJDK, but are merely hosted/enabled by the AdoptOpenJDK project as explicitly vanilla builds of the upstream OpenJDK project blessed/supported by the 8u/11u project lead.  For context, see https://github.com/docker-library/openjdk/issues/320#issuecomment-494050246:

> I am the OpenJDK 8 and 11 Updates OpenJDK project lead.

Concrete (potentially breaking) changes that are a consequence of this:

1. no more OpenJDK 7 images (see https://github.com/docker-library/openjdk/issues/320#issuecomment-494417274 -- "7 is still alive but only receives security updates" and "compatibility between 7 and 8 seems to be good enough in most cases")

2. no more OpenJDK 11 JRE images (see https://github.com/docker-library/openjdk/issues/320#issuecomment-494417274 -- "JDK 11 has no official JRE build. The reason for this is that the `jlink` tool allows users to create precisely-targeted stripped JDK images for their applications, and this is IMO better for Docker than a JRE image.")

3. the way OpenJDK is installed in 8 and 11 images is now quite different, and there will be some breakage expected -- I've tried to minimize that as much as possible, but there's only so much I can do here

4. no more OpenJDK 8 Alpine images (Alpine/musl is not officially supported by the OpenJDK project, so this reflects that -- see "Project Portola" for the Alpine porting efforts which I understand are still in need of help)

5. supported architectures are now decreased to match what's "officially" published -- for 8, we drop to just amd64, for 11 we drop to amd64 and arm64v8

@theRealAph if you'd be willing to at least take a cursory look over this, I'd really appreciate it :heart: (probably focused on either the `Dockerfile-adopt-*.template` files or one specific JDK version -- they're all generated from that same content so there's no reason to actually hand-review them all :sweat_smile:)

For extra visibility, explicit CC to the maintainers of other official images which are `FROM openjdk`:
- clojure: @Quantisan, @cap10morgan, @DeLaGuardo
- convertigo: @nicolas-albert, @opicciotto
- flink: @patricklucas, @iemejia
- geonetwork: @doublebyte1
- gradle: @keeganwitt
- groovy: @keeganwitt
- jetty: @md5, @gregw
- jruby: @cpuguy83
- kaazing-gateway:  @kaazing
- lightstreamer: @lightstreamer
- maven: @carlossg
- neo4j: @benbc, @digitalstain
- nuxeo: @dmetzler, @akervern
- orientdb: @luigidellaquila
- rapidoid: @nmihajlovski
- solr: @makuk66, @shalinmangar
- sonarqube: @janos-ss, @julienlancelot, @christophelevis
- storm: @31z4
- tomcat: @tianon, @yosifkit
- tomee: @lordofthejars, @otaviojava
- xwiki: @xwiki
- zookeeper: @31z4

I've personally looked at `tomcat`, and the practical implications there are going to be that we switch from being only JRE variants to being only JDK variants (since we need the JDK to compile Tomcat Native and it wasn't worth doing the legwork to make that work just for OpenJDK 8 JRE, and `jlink` doesn't work there because `tomcat` is intended to host arbitrary user code).

I also think it's worthwhile to point out the following other existing (and in-progress) JDK official images:

- https://hub.docker.com/_/amazoncorretto
- https://hub.docker.com/_/ibmjava
- https://github.com/docker-library/official-images/pull/5710 (`adoptopenjdk` proper)

Closes #309
Closes #303
Closes #244
Closes #189
Closes #278
Closes #217
Closes #73


tianon:
<details>
<summary> `bashbrew list` Diff: </summary>

```diff
--- /dev/fd/63	2019-05-22 16:26:01.809544532 -0700
+++ /dev/fd/62	2019-05-22 16:26:01.809544532 -0700
@@ -202,111 +202,129 @@
 openjdk:11.0-slim
 openjdk:11-jdk-slim
 openjdk:11-slim
-openjdk:11.0.3-jre-stretch
-openjdk:11.0-jre-stretch
-openjdk:11-jre-stretch
-openjdk:11.0.3-jre
-openjdk:11.0-jre
-openjdk:11-jre
-openjdk:11.0.3-jre-slim-stretch
-openjdk:11.0-jre-slim-stretch
-openjdk:11-jre-slim-stretch
-openjdk:11.0.3-jre-slim
-openjdk:11.0-jre-slim
-openjdk:11-jre-slim
+openjdk:11.0.3-jdk-windowsservercore-1809
+openjdk:11.0.3-windowsservercore-1809
+openjdk:11.0-jdk-windowsservercore-1809
+openjdk:11.0-windowsservercore-1809
+openjdk:11-jdk-windowsservercore-1809
+openjdk:11-windowsservercore-1809
+openjdk:11.0.3-jdk-windowsservercore
+openjdk:11.0.3-windowsservercore
+openjdk:11.0-jdk-windowsservercore
+openjdk:11.0-windowsservercore
+openjdk:11-jdk-windowsservercore
+openjdk:11-windowsservercore
+openjdk:11.0.3-jdk-windowsservercore-1803
+openjdk:11.0.3-windowsservercore-1803
+openjdk:11.0-jdk-windowsservercore-1803
+openjdk:11.0-windowsservercore-1803
+openjdk:11-jdk-windowsservercore-1803
+openjdk:11-windowsservercore-1803
+openjdk:11.0.3-jdk-windowsservercore
+openjdk:11.0.3-windowsservercore
+openjdk:11.0-jdk-windowsservercore
+openjdk:11.0-windowsservercore
+openjdk:11-jdk-windowsservercore
+openjdk:11-windowsservercore
+openjdk:11.0.3-jdk-windowsservercore-ltsc2016
+openjdk:11.0.3-windowsservercore-ltsc2016
+openjdk:11.0-jdk-windowsservercore-ltsc2016
+openjdk:11.0-windowsservercore-ltsc2016
+openjdk:11-jdk-windowsservercore-ltsc2016
+openjdk:11-windowsservercore-ltsc2016
+openjdk:11.0.3-jdk-windowsservercore
+openjdk:11.0.3-windowsservercore
+openjdk:11.0-jdk-windowsservercore
+openjdk:11.0-windowsservercore
+openjdk:11-jdk-windowsservercore
+openjdk:11-windowsservercore
+openjdk:8u212-b04-jdk-stretch
+openjdk:8u212-b04-stretch
 openjdk:8u212-jdk-stretch
 openjdk:8u212-stretch
 openjdk:8-jdk-stretch
 openjdk:8-stretch
+openjdk:8u212-b04-jdk
+openjdk:8u212-b04
 openjdk:8u212-jdk
 openjdk:8u212
 openjdk:8-jdk
 openjdk:8
+openjdk:8u212-b04-jdk-slim-stretch
+openjdk:8u212-b04-slim-stretch
 openjdk:8u212-jdk-slim-stretch
 openjdk:8u212-slim-stretch
 openjdk:8-jdk-slim-stretch
 openjdk:8-slim-stretch
+openjdk:8u212-b04-jdk-slim
+openjdk:8u212-b04-slim
 openjdk:8u212-jdk-slim
 openjdk:8u212-slim
 openjdk:8-jdk-slim
 openjdk:8-slim
-openjdk:8u212-jdk-alpine3.9
-openjdk:8u212-alpine3.9
-openjdk:8-jdk-alpine3.9
-openjdk:8-alpine3.9
-openjdk:8u212-jdk-alpine
-openjdk:8u212-alpine
-openjdk:8-jdk-alpine
-openjdk:8-alpine
+openjdk:8u212-b04-jdk-windowsservercore-1809
+openjdk:8u212-b04-windowsservercore-1809
 openjdk:8u212-jdk-windowsservercore-1809
 openjdk:8u212-windowsservercore-1809
 openjdk:8-jdk-windowsservercore-1809
 openjdk:8-windowsservercore-1809
+openjdk:8u212-b04-jdk-windowsservercore
+openjdk:8u212-b04-windowsservercore
 openjdk:8u212-jdk-windowsservercore
 openjdk:8u212-windowsservercore
 openjdk:8-jdk-windowsservercore
 openjdk:8-windowsservercore
+openjdk:8u212-b04-jdk-windowsservercore-1803
+openjdk:8u212-b04-windowsservercore-1803
 openjdk:8u212-jdk-windowsservercore-1803
 openjdk:8u212-windowsservercore-1803
 openjdk:8-jdk-windowsservercore-1803
 openjdk:8-windowsservercore-1803
+openjdk:8u212-b04-jdk-windowsservercore
+openjdk:8u212-b04-windowsservercore
 openjdk:8u212-jdk-windowsservercore
 openjdk:8u212-windowsservercore
 openjdk:8-jdk-windowsservercore
 openjdk:8-windowsservercore
+openjdk:8u212-b04-jdk-windowsservercore-ltsc2016
+openjdk:8u212-b04-windowsservercore-ltsc2016
 openjdk:8u212-jdk-windowsservercore-ltsc2016
 openjdk:8u212-windowsservercore-ltsc2016
 openjdk:8-jdk-windowsservercore-ltsc2016
 openjdk:8-windowsservercore-ltsc2016
+openjdk:8u212-b04-jdk-windowsservercore
+openjdk:8u212-b04-windowsservercore
 openjdk:8u212-jdk-windowsservercore
 openjdk:8u212-windowsservercore
 openjdk:8-jdk-windowsservercore
 openjdk:8-windowsservercore
+openjdk:8u212-b04-jre-stretch
 openjdk:8u212-jre-stretch
 openjdk:8-jre-stretch
+openjdk:8u212-b04-jre
 openjdk:8u212-jre
 openjdk:8-jre
+openjdk:8u212-b04-jre-slim-stretch
 openjdk:8u212-jre-slim-stretch
 openjdk:8-jre-slim-stretch
+openjdk:8u212-b04-jre-slim
 openjdk:8u212-jre-slim
 openjdk:8-jre-slim
-openjdk:8u212-jre-alpine3.9
-openjdk:8-jre-alpine3.9
-openjdk:8u212-jre-alpine
-openjdk:8-jre-alpine
-openjdk:7u221-jdk-jessie
-openjdk:7u221-jessie
-openjdk:7-jdk-jessie
-openjdk:7-jessie
-openjdk:7u221-jdk
-openjdk:7u221
-openjdk:7-jdk
-openjdk:7
-openjdk:7u221-jdk-slim-jessie
-openjdk:7u221-slim-jessie
-openjdk:7-jdk-slim-jessie
-openjdk:7-slim-jessie
-openjdk:7u221-jdk-slim
-openjdk:7u221-slim
-openjdk:7-jdk-slim
-openjdk:7-slim
-openjdk:7u211-jdk-alpine3.9
-openjdk:7u211-alpine3.9
-openjdk:7-jdk-alpine3.9
-openjdk:7-alpine3.9
-openjdk:7u211-jdk-alpine
-openjdk:7u211-alpine
-openjdk:7-jdk-alpine
-openjdk:7-alpine
-openjdk:7u221-jre-jessie
-openjdk:7-jre-jessie
-openjdk:7u221-jre
-openjdk:7-jre
-openjdk:7u221-jre-slim-jessie
-openjdk:7-jre-slim-jessie
-openjdk:7u221-jre-slim
-openjdk:7-jre-slim
-openjdk:7u211-jre-alpine3.9
-openjdk:7-jre-alpine3.9
-openjdk:7u211-jre-alpine
-openjdk:7-jre-alpine
+openjdk:8u212-b04-jre-windowsservercore-1809
+openjdk:8u212-jre-windowsservercore-1809
+openjdk:8-jre-windowsservercore-1809
+openjdk:8u212-b04-jre-windowsservercore
+openjdk:8u212-jre-windowsservercore
+openjdk:8-jre-windowsservercore
+openjdk:8u212-b04-jre-windowsservercore-1803
+openjdk:8u212-jre-windowsservercore-1803
+openjdk:8-jre-windowsservercore-1803
+openjdk:8u212-b04-jre-windowsservercore
+openjdk:8u212-jre-windowsservercore
+openjdk:8-jre-windowsservercore
+openjdk:8u212-b04-jre-windowsservercore-ltsc2016
+openjdk:8u212-jre-windowsservercore-ltsc2016
+openjdk:8-jre-windowsservercore-ltsc2016
+openjdk:8u212-b04-jre-windowsservercore
+openjdk:8u212-jre-windowsservercore
+openjdk:8-jre-windowsservercore
````

</details>

<details>
<summary> `bashbrew cat` Diff: </summary>

```diff
--- /dev/fd/63	2019-05-22 16:26:26.168867070 -0700
+++ /dev/fd/62	2019-05-22 16:26:26.168867070 -0700
@@ -59,107 +59,92 @@
 
 Tags: 11.0.3-jdk-stretch, 11.0.3-stretch, 11.0-jdk-stretch, 11.0-stretch, 11-jdk-stretch, 11-stretch
 SharedTags: 11.0.3-jdk, 11.0.3, 11.0-jdk, 11.0, 11-jdk, 11
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: 178c542fbb93a8f8a42e331b73a1214c9d8ba81d
+Architectures: amd64, arm64v8
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 11/jdk
 
 Tags: 11.0.3-jdk-slim-stretch, 11.0.3-slim-stretch, 11.0-jdk-slim-stretch, 11.0-slim-stretch, 11-jdk-slim-stretch, 11-slim-stretch, 11.0.3-jdk-slim, 11.0.3-slim, 11.0-jdk-slim, 11.0-slim, 11-jdk-slim, 11-slim
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: 178c542fbb93a8f8a42e331b73a1214c9d8ba81d
+Architectures: amd64, arm64v8
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 11/jdk/slim
 
-Tags: 11.0.3-jre-stretch, 11.0-jre-stretch, 11-jre-stretch
-SharedTags: 11.0.3-jre, 11.0-jre, 11-jre
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: 178c542fbb93a8f8a42e331b73a1214c9d8ba81d
-Directory: 11/jre
-
-Tags: 11.0.3-jre-slim-stretch, 11.0-jre-slim-stretch, 11-jre-slim-stretch, 11.0.3-jre-slim, 11.0-jre-slim, 11-jre-slim
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: 178c542fbb93a8f8a42e331b73a1214c9d8ba81d
-Directory: 11/jre/slim
-
-Tags: 8u212-jdk-stretch, 8u212-stretch, 8-jdk-stretch, 8-stretch
-SharedTags: 8u212-jdk, 8u212, 8-jdk, 8
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: b8ce9eff38451de3282b2eb2bcd8b520fb95e1ce
+Tags: 11.0.3-jdk-windowsservercore-1809, 11.0.3-windowsservercore-1809, 11.0-jdk-windowsservercore-1809, 11.0-windowsservercore-1809, 11-jdk-windowsservercore-1809, 11-windowsservercore-1809
+SharedTags: 11.0.3-jdk-windowsservercore, 11.0.3-windowsservercore, 11.0-jdk-windowsservercore, 11.0-windowsservercore, 11-jdk-windowsservercore, 11-windowsservercore
+Architectures: windows-amd64
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
+Directory: 11/jdk/windows/windowsservercore-1809
+Constraints: windowsservercore-1809
+
+Tags: 11.0.3-jdk-windowsservercore-1803, 11.0.3-windowsservercore-1803, 11.0-jdk-windowsservercore-1803, 11.0-windowsservercore-1803, 11-jdk-windowsservercore-1803, 11-windowsservercore-1803
+SharedTags: 11.0.3-jdk-windowsservercore, 11.0.3-windowsservercore, 11.0-jdk-windowsservercore, 11.0-windowsservercore, 11-jdk-windowsservercore, 11-windowsservercore
+Architectures: windows-amd64
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
+Directory: 11/jdk/windows/windowsservercore-1803
+Constraints: windowsservercore-1803
+
+Tags: 11.0.3-jdk-windowsservercore-ltsc2016, 11.0.3-windowsservercore-ltsc2016, 11.0-jdk-windowsservercore-ltsc2016, 11.0-windowsservercore-ltsc2016, 11-jdk-windowsservercore-ltsc2016, 11-windowsservercore-ltsc2016
+SharedTags: 11.0.3-jdk-windowsservercore, 11.0.3-windowsservercore, 11.0-jdk-windowsservercore, 11.0-windowsservercore, 11-jdk-windowsservercore, 11-windowsservercore
+Architectures: windows-amd64
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
+Directory: 11/jdk/windows/windowsservercore-ltsc2016
+Constraints: windowsservercore-ltsc2016
+
+Tags: 8u212-b04-jdk-stretch, 8u212-b04-stretch, 8u212-jdk-stretch, 8u212-stretch, 8-jdk-stretch, 8-stretch
+SharedTags: 8u212-b04-jdk, 8u212-b04, 8u212-jdk, 8u212, 8-jdk, 8
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 8/jdk
 
-Tags: 8u212-jdk-slim-stretch, 8u212-slim-stretch, 8-jdk-slim-stretch, 8-slim-stretch, 8u212-jdk-slim, 8u212-slim, 8-jdk-slim, 8-slim
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: b8ce9eff38451de3282b2eb2bcd8b520fb95e1ce
+Tags: 8u212-b04-jdk-slim-stretch, 8u212-b04-slim-stretch, 8u212-jdk-slim-stretch, 8u212-slim-stretch, 8-jdk-slim-stretch, 8-slim-stretch, 8u212-b04-jdk-slim, 8u212-b04-slim, 8u212-jdk-slim, 8u212-slim, 8-jdk-slim, 8-slim
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 8/jdk/slim
 
-Tags: 8u212-jdk-alpine3.9, 8u212-alpine3.9, 8-jdk-alpine3.9, 8-alpine3.9, 8u212-jdk-alpine, 8u212-alpine, 8-jdk-alpine, 8-alpine
-Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: ec1553cccfb87c5f53a38555771fa6d13cebfcba
-Directory: 8/jdk/alpine
-
-Tags: 8u212-jdk-windowsservercore-1809, 8u212-windowsservercore-1809, 8-jdk-windowsservercore-1809, 8-windowsservercore-1809
-SharedTags: 8u212-jdk-windowsservercore, 8u212-windowsservercore, 8-jdk-windowsservercore, 8-windowsservercore
+Tags: 8u212-b04-jdk-windowsservercore-1809, 8u212-b04-windowsservercore-1809, 8u212-jdk-windowsservercore-1809, 8u212-windowsservercore-1809, 8-jdk-windowsservercore-1809, 8-windowsservercore-1809
+SharedTags: 8u212-b04-jdk-windowsservercore, 8u212-b04-windowsservercore, 8u212-jdk-windowsservercore, 8u212-windowsservercore, 8-jdk-windowsservercore, 8-windowsservercore
 Architectures: windows-amd64
-GitCommit: d617d7e4b951735c512f3d5ac7672f7394115c91
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 8/jdk/windows/windowsservercore-1809
 Constraints: windowsservercore-1809
 
-Tags: 8u212-jdk-windowsservercore-1803, 8u212-windowsservercore-1803, 8-jdk-windowsservercore-1803, 8-windowsservercore-1803
-SharedTags: 8u212-jdk-windowsservercore, 8u212-windowsservercore, 8-jdk-windowsservercore, 8-windowsservercore
+Tags: 8u212-b04-jdk-windowsservercore-1803, 8u212-b04-windowsservercore-1803, 8u212-jdk-windowsservercore-1803, 8u212-windowsservercore-1803, 8-jdk-windowsservercore-1803, 8-windowsservercore-1803
+SharedTags: 8u212-b04-jdk-windowsservercore, 8u212-b04-windowsservercore, 8u212-jdk-windowsservercore, 8u212-windowsservercore, 8-jdk-windowsservercore, 8-windowsservercore
 Architectures: windows-amd64
-GitCommit: d617d7e4b951735c512f3d5ac7672f7394115c91
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 8/jdk/windows/windowsservercore-1803
 Constraints: windowsservercore-1803
 
-Tags: 8u212-jdk-windowsservercore-ltsc2016, 8u212-windowsservercore-ltsc2016, 8-jdk-windowsservercore-ltsc2016, 8-windowsservercore-ltsc2016
-SharedTags: 8u212-jdk-windowsservercore, 8u212-windowsservercore, 8-jdk-windowsservercore, 8-windowsservercore
+Tags: 8u212-b04-jdk-windowsservercore-ltsc2016, 8u212-b04-windowsservercore-ltsc2016, 8u212-jdk-windowsservercore-ltsc2016, 8u212-windowsservercore-ltsc2016, 8-jdk-windowsservercore-ltsc2016, 8-windowsservercore-ltsc2016
+SharedTags: 8u212-b04-jdk-windowsservercore, 8u212-b04-windowsservercore, 8u212-jdk-windowsservercore, 8u212-windowsservercore, 8-jdk-windowsservercore, 8-windowsservercore
 Architectures: windows-amd64
-GitCommit: d617d7e4b951735c512f3d5ac7672f7394115c91
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 8/jdk/windows/windowsservercore-ltsc2016
 Constraints: windowsservercore-ltsc2016
 
-Tags: 8u212-jre-stretch, 8-jre-stretch
-SharedTags: 8u212-jre, 8-jre
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: b8ce9eff38451de3282b2eb2bcd8b520fb95e1ce
+Tags: 8u212-b04-jre-stretch, 8u212-jre-stretch, 8-jre-stretch
+SharedTags: 8u212-b04-jre, 8u212-jre, 8-jre
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 8/jre
 
-Tags: 8u212-jre-slim-stretch, 8-jre-slim-stretch, 8u212-jre-slim, 8-jre-slim
-Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: b8ce9eff38451de3282b2eb2bcd8b520fb95e1ce
+Tags: 8u212-b04-jre-slim-stretch, 8u212-jre-slim-stretch, 8-jre-slim-stretch, 8u212-b04-jre-slim, 8u212-jre-slim, 8-jre-slim
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
 Directory: 8/jre/slim
 
-Tags: 8u212-jre-alpine3.9, 8-jre-alpine3.9, 8u212-jre-alpine, 8-jre-alpine
-Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: ec1553cccfb87c5f53a38555771fa6d13cebfcba
-Directory: 8/jre/alpine
-
-Tags: 7u221-jdk-jessie, 7u221-jessie, 7-jdk-jessie, 7-jessie
-SharedTags: 7u221-jdk, 7u221, 7-jdk, 7
-Architectures: amd64, arm32v5, arm32v7, i386
-GitCommit: 8259a15a9118fbe624e47979d34fac3ea7059ac4
-Directory: 7/jdk
-
-Tags: 7u221-jdk-slim-jessie, 7u221-slim-jessie, 7-jdk-slim-jessie, 7-slim-jessie, 7u221-jdk-slim, 7u221-slim, 7-jdk-slim, 7-slim
-Architectures: amd64, arm32v5, arm32v7, i386
-GitCommit: 8259a15a9118fbe624e47979d34fac3ea7059ac4
-Directory: 7/jdk/slim
-
-Tags: 7u211-jdk-alpine3.9, 7u211-alpine3.9, 7-jdk-alpine3.9, 7-alpine3.9, 7u211-jdk-alpine, 7u211-alpine, 7-jdk-alpine, 7-alpine
-Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: bc2fcfad89194727ec8a9334730af6ff4d8c44a7
-Directory: 7/jdk/alpine
-
-Tags: 7u221-jre-jessie, 7-jre-jessie
-SharedTags: 7u221-jre, 7-jre
-Architectures: amd64, arm32v5, arm32v7, i386
-GitCommit: 8259a15a9118fbe624e47979d34fac3ea7059ac4
-Directory: 7/jre
-
-Tags: 7u221-jre-slim-jessie, 7-jre-slim-jessie, 7u221-jre-slim, 7-jre-slim
-Architectures: amd64, arm32v5, arm32v7, i386
-GitCommit: 8259a15a9118fbe624e47979d34fac3ea7059ac4
-Directory: 7/jre/slim
-
-Tags: 7u211-jre-alpine3.9, 7-jre-alpine3.9, 7u211-jre-alpine, 7-jre-alpine
-Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
-GitCommit: bc2fcfad89194727ec8a9334730af6ff4d8c44a7
-Directory: 7/jre/alpine
+Tags: 8u212-b04-jre-windowsservercore-1809, 8u212-jre-windowsservercore-1809, 8-jre-windowsservercore-1809
+SharedTags: 8u212-b04-jre-windowsservercore, 8u212-jre-windowsservercore, 8-jre-windowsservercore
+Architectures: windows-amd64
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
+Directory: 8/jre/windows/windowsservercore-1809
+Constraints: windowsservercore-1809
+
+Tags: 8u212-b04-jre-windowsservercore-1803, 8u212-jre-windowsservercore-1803, 8-jre-windowsservercore-1803
+SharedTags: 8u212-b04-jre-windowsservercore, 8u212-jre-windowsservercore, 8-jre-windowsservercore
+Architectures: windows-amd64
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
+Directory: 8/jre/windows/windowsservercore-1803
+Constraints: windowsservercore-1803
+
+Tags: 8u212-b04-jre-windowsservercore-ltsc2016, 8u212-jre-windowsservercore-ltsc2016, 8-jre-windowsservercore-ltsc2016
+SharedTags: 8u212-b04-jre-windowsservercore, 8u212-jre-windowsservercore, 8-jre-windowsservercore
+Architectures: windows-amd64
+GitCommit: cb091f035299421cc58ff8ea5126bb8a9350da97
+Directory: 8/jre/windows/windowsservercore-ltsc2016
+Constraints: windowsservercore-ltsc2016
````

</details>

**Comment by tianon** [comment]
As a later issue, I'd love to also publish pre-release images based on the "EA" builds from https://adoptopenjdk.net/upstream.html, but that's a larger project and I don't want to hold this up.

**Comment by conn** [comment]
CC @carlossg and @ndeloof for [Jenkins](https://github.com/jenkinsci/docker)

And also pointing to [Zulu's JDK image](https://github.com/zulu-openjdk/zulu-openjdk)

**Comment by theRealAph** [comment]
> Closes #320
> 
> > These binaries are built by Red Hat on behalf of the OpenJDK jdk8u and jdk11u projects. The binaries are created from the unmodified source code at OpenJDK.
> 
> So to put that another way and be explicitly clear, the builds from https://adoptopenjdk.net/upstream.html are _not_ AdoptOpenJDK, but are merely hosted/enabled by the AdoptOpenJDK project as explicitly vanilla builds of the upstream OpenJDK project blessed/supported by the 8u/11u project lead. For context, see [#320 (comment)](https://github.com/docker-library/openjdk/issues/320#issuecomment-494050246):
> 
> > I am the OpenJDK 8 and 11 Updates OpenJDK project lead.
> 
> Concrete (potentially breaking) changes that are a consequence of this:
> 
>     1. no more OpenJDK 7 images (see [#320 (comment)](https://github.com/docker-library/openjdk/issues/320#issuecomment-494417274) -- "7 is still alive but only receives security updates" and "compatibility between 7 and 8 seems to be good enough in most cases")

Yes. Red Hat long-term support ends July 2020, and I expect other vendors to do something similar. Over time it becomes hard and harder to back-port fixes.

>     2. no more OpenJDK 11 JRE images (see [#320 (comment)](https://github.com/docker-library/openjdk/issues/320#issuecomment-494417274) -- "JDK 11 has no official JRE build. The reason for this is that the `jlink` tool allows users to create precisely-targeted stripped JDK images for their applications, and this is IMO better for Docker than a JRE image.")

Right, for now.

>     3. the way OpenJDK is installed in 8 and 11 images is now quite different, and there will be some breakage expected -- I've tried to minimize that as much as possible, but there's only so much I can do here
>
>     4. no more OpenJDK 8 Alpine images (Alpine/musl is not officially supported by the OpenJDK project, so this reflects that -- see "Project Portola" for the Alpine porting efforts which I understand are still in need of help)
> 
>     5. supported architectures are now decreased to match what's "officially" published -- for 8, we drop to just amd64, for 11 we drop to amd64 and arm64v8

What other architectures do you need?

File: .appveyor.yml [hunk 1]
```diff
@@ -7,6 +7,10 @@ environment:
       variant: windowsservercore-ltsc2016
     - version: 12
       variant: windowsservercore-ltsc2016
+    - version: 11
+      variant: windowsservercore-ltsc2016
+    - version: 8
```
shipilev: Isn't this line exactly like the lines below it?

Dockerfile-adopt-slim.template
```diff
@@ -0,0 +1,60 @@
...
+		gnupg dirmngr \
+	; \
+	rm -rf /var/lib/apt/lists/*; \
+	\
+	wget -O openjdk.tgz.asc "${JAVA_BASE_URL}${upstreamArch}_linux_${JAVA_URL_VERSION}.tar.gz.sign"; \
+	wget -O openjdk.tgz "${JAVA_BASE_URL}${upstreamArch}_linux_${JAVA_URL_VERSION}.tar.gz" --progress=dot:giga; \
+	\
+	export GNUPGHOME="$(mktemp -d)"; \
+# TODO find a good link for users to verify this key is right (https://mail.openjdk.java.net/pipermail/jdk-updates-dev/2019-April/000951.html is one of the only mentions of it I can find); perhaps a note added to https://adoptopenjdk.net/upstream.html would make sense?
+	gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys CA5F11C6CE22644D42C6AC4492EF8D39DC13168F; \
```
jerboaa: The binaries we release are not only signed with the Red Hat OpenJDK gpg key, but that key is also signed by @theRealAph. As such, I'd encourage to use those additional steps before verifying the binary itself:

```
gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys 0x332FA671
gpg --batch --list-sig 0x9FF32E37 | grep 'Andrew Haley'
gpg --batch --verify ...
```

8/jdk/Dockerfile
```diff
@@ -1,69 +1,59 @@
...
-RUN ln -svT "/usr/lib/jvm/java-8-openjdk-$(dpkg --print-architecture)" /docker-java-home
-ENV JAVA_HOME /docker-java-home
+# backwards compatibility shim
+RUN { echo '#/bin/sh'; echo 'echo "$JAVA_HOME"'; } > /usr/local/bin/docker-java-home && chmod +x /usr/local/bin/docker-java-home && [ "$JAVA_HOME" = "$(docker-java-home)" ]
 
-ENV JAVA_VERSION 8u212
-ENV JAVA_DEBIAN_VERSION 8u212-b01-1~deb9u1
+# https://adoptopenjdk.net/upstream.html
+ENV JAVA_VERSION 8u212-b04
+ENV JAVA_BASE_URL https://github.com/AdoptOpenJDK/openjdk8-upstream-binaries/releases/download/jdk8u212-b04/OpenJDK8U-
```
jerboaa: FYI: Going forward (with the July CPU) the JDK URLs will change to:

`https://github.com/AdoptOpenJDK/openjdk8-upstream-binaries/releases/download/<tag>/OpenJDK8U-jdk_`

That is, they'll be similar to JRE links.

tianon:
Want to start by saying again how grateful we are for the review/help. :+1: :heart:

(I'll start with this question then hit into the code review comments! :tada:)

> What other architectures do you need?

Well, I think "need" is a bit of a strong way to put it. :sweat_smile: 

We currently have the intersection of every architecture Debian's packages supported and that the official images program supports, which is `amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x`.  I don't think we necessarily *need* all of those, especially with https://github.com/docker-library/official-images/pull/5710 having so many more, but it is a pretty drastic change to go from 7 down to 2 or even 1 (although back +1 on 11 thanks to Windows :metal:).

I know IBM has been keen on ppc64le and s390x in the past, but they're shifting their focus almost entirely to AdoptOpenJDK (as I've understood), so those likely aren't going to be a very heavy loss here (as much as I personally enjoy them :smile:).

Arm64 (which exists here for 11) is definitely really important to Arm, so it would be appreciated if that could stick around.  It would be nice if we could have Arm32 (v7) at least, but I'm not going to throw any fits (and am really quite impressed/thrilled that we've got arm64 on 11 :heart:).

Also, I want to be clear that we don't expect you to release anything that's not fully "supported" within the OpenJDK project yet (Alpine / Portola being a great example), and contrary to some of the opinions shared in that previous thread we really do want to represent OpenJDK as well as we can. :sweat_smile: :blue_heart:

.appveyor.yml
tianon: Yesss, thanks for noticing. :innocent:

I refactored the script that generated these lines and the part that spits out this was inside a loop that it wasn't part of previously so it got double-added. :laughing:

Dockerfile-adopt-slim.template
tianon: Oh neat!  Any chance of adding that info/fingerprint to https://adoptopenjdk.net/upstream.html? :innocent:

~~There's not a way to tell `--verify` to look for both keys, right?  (Short of doing something like `--status-fd` and parsing the "machine readable" output it creates?)
I'm happy to add the additional key, but if I just add it now it'll just be verifying one or the other AFAIK.~~

Wrote all that and then realized what you mean -- you mean the *RedHat* key is signed by Andrew (not the binary releases). :man_facepalming: :innocent: :heart:   I'll take a look! (We don't typically worry about signatures in the official images since most of our upstreams publish a full fingerprint anyhow and that's what we figure users are going to go check/verify if they're going to do anything, but I do agree it makes some sense here, especially for folks who go, "wait signed by RedHat? what's this?" :+1:)

8/jdk/Dockerfile
tianon: Oh nice! :+1:

Any chance I could convince you to publish something like https://adoptopenjdk.net/upstream.json with some useful deets so I'm not scraping HTML over in `update.sh` to get these values? :innocent:

(Published checksums would be cool too, but I don't want to be going overboard and fully understand if I'm asking too much on either of these. :sweat_smile:)

nicolas-albert:
Hi,
I try to understand the implication for my `convertigo` official image.
I didn't directly depend on the `openjdk` image but from `tomcat:7-jre8` and `tomcat:7-jre8-alpine`.
OpenJDK 8 is still available. Do you keep the JRE or only the JDK (not a problem for us) ?
Do I have to drop some arch (`amd64`, `arm32v6`, `arm64v8`, `i386`) ?
Thanks !

tianon:
@nicolas-albert the short is that `tomcat:*-alpine` will go away, and `tomcat:7-jre8` will turn into `tomcat:7-jdk8` (for now -- hopefully we can find a reasonable solution for building Tomcat Native in JRE images in the future like using `COPY --from` via https://github.com/docker-library/official-images/pull/5929, but for now that's not part of the plan given that it wouldn't be supported downstream yet)

Since you're looking at `jre8`, your architectures list will unfortunately drop just to just `amd64` (unless we can convince these nice OpenJDK upstream folks that any of those other architectures can be published for 8 :innocent:).

8/jdk/Dockerfile
tianon: Ah, just found https://github.com/AdoptOpenJDK/openjdk-website/blob/master/src/handlebars/upstream.handlebars (so a `.json` is probably asking a bit much) -- maybe it would be possible to get these builds integrated with https://api.adoptopenjdk.net/ somehow?

Not a huge problem, we can adapt as necessary (it usually becomes pretty obvious when our scraping breaks since we do build tests afterwards before doing any automated commits). :+1:

jerboaa: Yes. The idea is to integrate this via https://api.adoptopenjdk.net/ long term.

yosifkit:
Just making a rough note of packages that might now be "missing" since they are dependencies of Debian's `-jdk`, `-jre` or `-headless` packages:

- openjdk-8-jdk:
    `libasound2`
    `libatk-wrapper-java-jni`
    `libgl1-mesa-glx`
    `libgtk-3-0`
    `libpng16-16`
    `libpulse0`
    `libx11-6`
    `libxext6`
    `libxinerama1`
    `libxrandr2`
- openjdk-11-jdk:
    `libasound2`
    `libcups2`
    `libgcc1`
    `libjpeg62-turbo`
    `liblcms2-2`
    `libnss3`
    `libpcsclite1`
    `libpng16-16`
    `libstdc++6`
    `libx11-6`
    `libxext6`
    `libxi6`
    `libxinerama1`
    `libxrandr2`
    `libxrender1`
    `libxtst6`
    `util-linux`

11/jdk/slim/Dockerfile
```diff
@@ -1,86 +1,63 @@
...
-#  2. Compiling OpenJDK also requires the JDK to be installed, and it gets
-#     really hairy.
-#
-#     For some sample build times, see Debian's buildd logs:
-#       https://buildd.debian.org/status/logs.php?pkg=openjdk-11
-
-RUN apt-get update && apt-get install -y --no-install-recommends \
-		bzip2 \
-		unzip \
-		xz-utils \
```
yosifkit: Note to future self: it seems this was left in the `slim` variant since the beginning by accident because they were just a copy of the regular with 2 line changes provided by a `sed`: https://github.com/docker-library/openjdk/pull/133/files#diff-25e6ad8ebc5d26ba06d2bcb0b220149cR351. I think we could reasonably consider adding `unzip` back for useful `.jar` extraction.

tianon:
For those unfamiliar with the official-images process: the image will be updated once https://github.com/docker-library/official-images/pull/5953 merges (which will include a diff and a build test). :+1:

J0WI:
Sad to see the Alpine variants gone :confused:

tianon:
Yeah, it's a shame, but I don't really see a better option (and it reflects the upstream OpenJDK support for Alpine accurately).

J0WI:
Why have they been removed entirely?
The Alpine builds are based on https://icedtea.classpath.org/ which is maintained by RedHat. And they even support more platforms than AdoptOpenJDK does.
There is https://github.com/docker-library/official-images/pull/5710 for AdoptOpenJDK, but replacing the sources here will break many containers and running systems.

tianon:
Because they were installing from distribution-supplied packages -- do you know if the TCK has been run on those, for example?  (That's the whole reason we got into this mess with Debian in the first place, so without some confirmation from OpenJDK folks that the Alpine packages are ~ok, I'd rather avoid keeping ourselves in this hole.)

J0WI:
https://icedtea.classpath.org/openjdk/java/faq.jsp.html

https://mail.openjdk.java.net/pipermail/jdk8u-dev/2019-May/009330.html doesn't mention Alpine or IcedTea.

tianon:
Sure, they're building from the IcedTea source, but Alpine doesn't package
official IcedTea builds either:
https://git.alpinelinux.org/aports/tree/community/openjdk8/APKBUILD?h=3.9-stable

The exact complaint doesn't apply to the Alpine builds, sure, but the
spirit of it (using unverified, unknown builds) certainly does, which is
why they were removed here as well.

J0WI:
> but the spirit of it (using unverified, unknown builds) certainly does, which is why they were removed here as well.

AdoptOpenJDK has the same issue among other disadvantages (multiarch, image footprint).

giltene:
> > but the spirit of it (using unverified, unknown builds) certainly does, which is why they were removed here as well.
> 
> AdoptOpenJDK has the same issue among other disadvantages (multiarch, image footprint).

The specific upstream builds **_hosted at_** https://adoptopenjdk.net/upstream.html are fully TCK tested to my knowledge, and signed by the OpenJDK Updates project lead for the associated project (@theRealAph). see https://mail.openjdk.java.net/pipermail/jdk8u-dev/2019-April/009105.html . Their pedigree is probably as good as you can ask for.

[And yes, it would be useful for the page containing them to make that clear without digging into the jdk8u-dev mailing list]

iemejia:
Have there been discussions with the adoptopenjdk community on they taking maintenance of these images or contributing other architectures support?
They already provide images and seem to support some of the now missing images, e.g. alpine, 11:jre.'
https://hub.docker.com/r/adoptopenjdk/openjdk11 and since we rely mostly in the same sources probably it is worth to do a join effort.

tianon:
@iemejia we're not using AdoptOpenJDK -- they just happen to be the host for the binaries we are using (which are 100% vanilla OpenJDK binaries compiled direct from the OpenJDK source as described on https://adoptopenjdk.net/upstream.html)

You'll likely be interested in https://github.com/docker-library/official-images/pull/5710 / https://hub.docker.com/_/adoptopenjdk which merged last week too and is an `adoptopenjdk` official image maintained by the AdoptOpenJDK community directly (cc @dinogun :+1: :heart:).

jdimmerman:
Was changing JAVA_HOME purposeful @tianon?

tianon:
Yes, the old value of that environment variable was a hack to work around the fact that Debian's packages for OpenJDK used an architecture-specific directory name and we included a script to update it based on whether the JDK was installed.  This includes an updated `JAVA_HOME` variable and still includes a dummy version of our old script for backwards compatibility.

jdimmerman:
ok very cool, thanks @tianon

keeganwitt:
It broke sonarqube chart also (https://github.com/helm/charts/issues/14136 / https://github.com/helm/charts/commit/916c1fd258182fae6f1f5187339972a839722fa0), probably multiple downstream stuff will be affected.

giltene:
Moving JAVA_HOME to something more openjdk-private like e.g. /usr/local-openjdk/openjdk-8 would probably help stay away from such downstream assumptions about being able to stomp  /use/local

keeganwitt:
Maybe, but that doesn't really follow the Filesystem Hierarchy Standard.  I do like that the new path is more compliant, but it's unfortunate that it's going to cause a number of breakages.  If we were to choose not to try to follow FHS, then we might as well install to the old path.

giltene:
@keeganwitt : Ok, fair point. So use something like /opt/openjdk/openjdk-8 that would be FHS compliant. I wouldn’t use the old location, as that one can get stomped by deb package installs.

tianon:
I would say `/usr/local` is an extremely generic path and *the* common default for non-distro-packaged software installation, so IMO it wasn't terribly sane to assume it was safe to unilaterally clobber even on the previous versions of the images where it happened to be ~safe -- even in those images we've had `/usr/local/bin/docker-java-home` installed under `/usr/local` for quite a long time now (since it was first introduced), they just happened to not need that script so didn't mind (or perhaps notice) it being clobbered as well.

Downstream breakage here is unfortunate (and really somewhat expected, given the nature of the change no matter how we resolved this) and we've done our best to ensure things like `JAVA_HOME` are set appropriately in an attempt to help mitigate, but there's only so much we can do -- hopefully the breakage is reasonably minor and/or easily updated and perhaps the breakage will help communicate that there has been a change that folks might be interested in looking into the details of.

jdimmerman:
Thanks @tianon. Understood.

Is there a good way to subscribe to or identify ahead of time breaking changes as a consumer of this repo (or, more specifically, the docker-hub distribution)?

tianon:
@jdimmerman for official images such as this one, the best we've got is GitHub, so I'd recommend either monitoring pull requests to the respective repositories of the images you're interested in (such as this one) or pull requests to the labels you're interested in on https://github.com/docker-library/official-images, such as https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fopenjdk (which will typically include a full diff of the changes but not always the same level of change explanation as the original PR to this repository)

jdimmerman:
Thanks @tianon

PR #328, opened 2019-05-28
Auto-update "cacerts" bundle via Debian's "ca-certificates" package

### LINKED ISSUE #327: cacerts not updated with `update-ca-certificates`
The usual way to add certificate to a debian system by copying them to `/usr/local/share/ca-certificates/` and running `update-ca-certificates` broke with the switch to AdoptOpenJDK.
We've tested with `openjdk:8` that was updated a few days ago (sha256:d318d6aeb831170ba394aae69e0c33cf4bcd4ab2a90ff60f50e68664fac09ecf).
Usually, `update-ca-certificates` updates `/etc/ssl/certs/java/cacerts` and this files is sim-linked in the jdk `jre/lib/security/cacerts`.
However since the switch to AdoptOpenJDK, it no longer happens.
We need to have a script like http://rabexc.org/posts/certificates-not-working-java#comment-4099504075 to update the JDK cacerts file.

tianon:
Closes #327

Many thanks to ncopa for the `trust` CLI which is *way* faster than doing this same thing via `keytool`. :sweat_smile:

### DIFF
--- a/11/jdk/Dockerfile
+++ b/11/jdk/Dockerfile
@@ -7,6 +7,9 @@ RUN set -eux; \
 		unzip \
 		xz-utils \
 		\
+# utilities for keeping Debian and OpenJDK CA certificates in sync
+		ca-certificates p11-kit \
+		\
 # java.lang.UnsatisfiedLinkError: /usr/local/openjdk-11/lib/libfontmanager.so: libfreetype.so.6: cannot open shared object file: No such file or directory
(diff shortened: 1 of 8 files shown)


tianon:
On `openjdk:11-jdk-slim`, this adjusts the image from ~377MB up to ~384MB which IMO is pretty acceptable given the benefits.

PR #348, opened 2019-08-14
Added support for Ubuntu 18.04, 16.04, and CentOS 7

apolloclark:
I added support for Ubuntu 18.04 Bionic LTS, Ubuntu 16.04 Xenial LTS, and CentOS 7. Luckily the Debian 10 Buster Dockerfile was the basis for the Ubuntu images with minimal changes. The CentOS 7 Dockerfile required converting the apt-get commands to yum, and changing the pathway for the CA Certs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,12 @@ matrix:
       env: VERSION=11
     - os: linux
       env: VERSION=11 VARIANT=slim
+    - os: linux
+      env: VERSION=11 VARIANT=ubuntu18.04
+    - os: linux
+      env: VERSION=11 VARIANT=ubuntu16.04
+    - os: linux
+      env: VERSION=11 VARIANT=centos7
     - os: windows
       dist: 1803-containers
       env: VERSION=11 VARIANT=windows/windowsservercore-1803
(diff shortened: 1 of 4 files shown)


yosifkit:
My answer is the same as my comment on https://github.com/docker-library/python/issues/61#issuecomment-282450114

apolloclark:
@yosifkit Thanks for the info. Yeah, I can see where Docker is coming from, it is really tedious to maintain multiple images with different base OSes. Yes, in theory no one should have to think about it.

In practice, in my consulting work with large enterprises that are now adopting Docker images, they have each standardized on a given base OS in their AWS / Azure / GCP / VMware environment, and want to use the same in Docker. It's usually either Ubuntu 18.04 / 16.04 or a mix of RHEL / CentOS. When teams of 100+ devs / ops / security / qa engineers need to debug the containers, it's really hard to do. The given base OSes use different dependencies, different dependency versions, folder pathways are different, and there are still base OS specific configs in things like Tomcat.

Analysis on the wide variety of base OSesof the most popular Docker images:
https://gist.github.com/apolloclark/8f339de76182d5fcebff7e5b3a24159c

I've built a Gradle + Packer + Ansible + Serverspec, build pipeline to rebuild some of those images across the other popular base OSes:
https://github.com/apolloclark/packer-docker#support-matrix

In an ideal world, every company / team / dev would pick a given base OS, and pull only images using it, and the image would be branded and supported as an "Official Docker Image", instead of pulling from untrusted random developers. Sorry for the rant, thanks for listening. 😁

tianon:
Unfortunately, this isn't something we're going to support/maintain -- thanks for the contribution, though! :+1:

PR #355, opened 2019-09-06
Add Debian variants of Oracle-consuming versions

### LINKED ISSUE #302: Where's the OpenJDK 12 image?
There are only oracle and windows images present for well, whereas for 11 there's an openjdk image and a slim image. When can we expect these for 12?

tianon:
These are images based on the official JDK artifacts for 12+ released on https://jdk.java.net/ with a Debian base in addition to Oracle.

Closes #302

<details>
<summary>A Useful Diff:</summary>

```diff
$ diff -u <(bashbrew cat openjdk) <(bashbrew cat <(./generate-stackbrew-library.sh))
--- /dev/fd/63	2019-09-06 12:13:35.609665720 -0700
+++ /dev/fd/62	2019-09-06 12:13:35.609665720 -0700
@@ -7,6 +7,14 @@
 Directory: 14/jdk/oracle
 Constraints: !aufs
 
+Tags: 14-ea-13-jdk-buster, 14-ea-13-buster, 14-ea-jdk-buster, 14-ea-buster, 14-jdk-buster, 14-buster
+GitCommit: b26a21dd82ee269efe26a9407eff9235bdde3dad
+Directory: 14/jdk
+
+Tags: 14-ea-13-jdk-slim-buster, 14-ea-13-slim-buster, 14-ea-jdk-slim-buster, 14-ea-slim-buster, 14-jdk-slim-buster, 14-slim-buster, 14-ea-13-jdk-slim, 14-ea-13-slim, 14-ea-jdk-slim, 14-ea-slim, 14-jdk-slim, 14-slim
+GitCommit: b26a21dd82ee269efe26a9407eff9235bdde3dad
+Directory: 14/jdk/slim
+
 Tags: 14-ea-12-jdk-alpine3.10, 14-ea-12-alpine3.10, 14-ea-jdk-alpine3.10, 14-ea-alpine3.10, 14-jdk-alpine3.10, 14-alpine3.10, 14-ea-12-jdk-alpine, 14-ea-12-alpine, 14-ea-jdk-alpine, 14-ea-alpine, 14-jdk-alpine, 14-alpine
 GitCommit: b57ab1457d190f313c46ffbec2b994b041b4d08c
 Directory: 14/jdk/alpine
@@ -38,6 +46,14 @@
 Directory: 13/jdk/oracle
 Constraints: !aufs
 
+Tags: 13-jdk-buster, 13-buster
+GitCommit: b26a21dd82ee269efe26a9407eff9235bdde3dad
+Directory: 13/jdk
+
+Tags: 13-jdk-slim-buster, 13-slim-buster, 13-jdk-slim, 13-slim
+GitCommit: b26a21dd82ee269efe26a9407eff9235bdde3dad
+Directory: 13/jdk/slim
+
 Tags: 13-jdk-windowsservercore-1809, 13-windowsservercore-1809
 SharedTags: 13-jdk-windowsservercore, 13-windowsservercore, 13-jdk, 13
 Architectures: windows-amd64
@@ -65,6 +81,14 @@
 Directory: 12/jdk/oracle
 Constraints: !aufs
 
+Tags: 12.0.2-jdk-buster, 12.0.2-buster, 12.0-jdk-buster, 12.0-buster, 12-jdk-buster, 12-buster, jdk-buster, buster
+GitCommit: b26a21dd82ee269efe26a9407eff9235bdde3dad
+Directory: 12/jdk
+
+Tags: 12.0.2-jdk-slim-buster, 12.0.2-slim-buster, 12.0-jdk-slim-buster, 12.0-slim-buster, 12-jdk-slim-buster, 12-slim-buster, jdk-slim-buster, slim-buster, 12.0.2-jdk-slim, 12.0.2-slim, 12.0-jdk-slim, 12.0-slim, 12-jdk-slim, 12-slim, jdk-slim, slim
+GitCommit: b26a21dd82ee269efe26a9407eff9235bdde3dad
+Directory: 12/jdk/slim
+
 Tags: 12.0.2-jdk-windowsservercore-1809, 12.0.2-windowsservercore-1809, 12.0-jdk-windowsservercore-1809, 12.0-windowsservercore-1809, 12-jdk-windowsservercore-1809, 12-windowsservercore-1809, jdk-windowsservercore-1809, windowsservercore-1809
 SharedTags: 12.0.2-jdk-windowsservercore, 12.0.2-windowsservercore, 12.0-jdk-windowsservercore, 12.0-windowsservercore, 12-jdk-windowsservercore, 12-windowsservercore, jdk-windowsservercore, windowsservercore, 12.0.2-jdk, 12.0.2, 12.0-jdk, 12.0, 12-jdk, 12, jdk, latest
 Architectures: windows-amd64
````

</details>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,20 @@ matrix:
     - os: linux
       env: VERSION=14 VARIANT=oracle
     - os: linux
+      env: VERSION=14
+    - os: linux
+      env: VERSION=14 VARIANT=slim
+    - os: linux
       env: VERSION=14 VARIANT=alpine
     - os: windows
       dist: 1803-containers
       env: VERSION=14 VARIANT=windows/windowsservercore-1803
     - os: linux
       env: VERSION=13 VARIANT=oracle
+    - os: linux
+      env: VERSION=13
+    - os: linux
+      env: VERSION=13 VARIANT=slim
     - os: windows
       dist: 1803-containers
       env: VERSION=13 VARIANT=windows/windowsservercore-1803
--- a/13/jdk/Dockerfile
+++ b/13/jdk/Dockerfile
@@ -1,0 +1,81 @@
+FROM buildpack-deps:buster-scm
+
+RUN set -eux; \
+	apt-get update; \
+	apt-get install -y --no-install-recommends \
+		bzip2 \
+		unzip \
+		xz-utils \
+		\
+# utilities for keeping Debian and OpenJDK CA certificates in sync
+		ca-certificates p11-kit \
+		\
+# jlink --strip-debug on 13+ needs objcopy: https://github.com/docker-library/openjdk/issues/351
+# Error: java.io.IOException: Cannot run program "objcopy": error=2, No such file or directory
+		binutils \
+# java.lang.UnsatisfiedLinkError: /usr/local/openjdk-11/lib/libfontmanager.so: libfreetype.so.6: cannot open shared object file: No such file or directory
+# java.lang.NoClassDefFoundError: Could not initialize class sun.awt.X11FontManager
+# https://github.com/docker-library/openjdk/pull/235#issuecomment-424466077
+		fontconfig libfreetype6 \
+	; \
+	rm -rf /var/lib/apt/lists/*
+
+# Default to UTF-8 file.encoding
+ENV LANG C.UTF-8
+
+ENV JAVA_HOME /usr/java/openjdk-13
+ENV PATH $JAVA_HOME/bin:$PATH
+
+# backwards compatibility shim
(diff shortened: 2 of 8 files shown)


tianon:
To my point in https://github.com/docker-library/openjdk/issues/302#issuecomment-482611053:

> I do want to be very clear that if we reintroduce Debian variants for 12+ they will not match the existing Debian variants, which is a big point to hesitate on here IMO.

I think we're reasonably safe to reintroduce these given the amount of time that has now passed since the old versions were deprecated (including a major Debian release, so these are Buster-based instead of Stretch-based).

yosifkit:
Should we add something like https://github.com/docker-library/php/pull/542?

tianon:
Oh, very interesting thought, although I think that's slightly orthogonal (given that we already have the Adopt variants for 8 and 11 which do not include something like that but could probably also benefit).

I think it's a little bit harder to achieve though given the varying nature of JRE/JDK package names (things like `default-jre` are a common dependency :frowning_face:).

tianon:
Rebased/updated :+1:

tianon:
... and now the relevant half of the CI is green :smile:

PR #359, opened 2019-09-25
Added support for windowsservercore 1903.

BlueAndi:


BlueAndi:
@yosifkit The travis build seems to fail, because the 1903 base image is not supported by travis. Shall I just remove them from the travis build?

yosifkit:
https://github.com/docker-library/golang/pull/289#issuecomment-509841894

PR #370, opened 2019-11-13
Added java env

brunomartinspro:
Running the image as "FROM openjdk:14-ea-22-jdk-windowsservercore-1809" does not create the "java" environment variable so it can run on other containers that use "java" instead of "JAVA_HOME"

### DIFF
--- a/14/jdk/windows/windowsservercore-1809/Dockerfile
+++ b/14/jdk/windows/windowsservercore-1809/Dockerfile
@@ -17,6 +17,7 @@ RUN Write-Host 'Enabling TLS 1.2 (https://githubengineering.com/crypto-removal-n
 	New-ItemProperty -Path ('{0}/Server' -f $tls12RegBase) -Name 'Enabled' -PropertyType DWORD -Value 1 -Force
 
 ENV JAVA_HOME C:\\openjdk-14
+ENV java C:\\openjdk-14\\bin\\java.exe
 RUN $newPath = ('{0}\bin;{1}' -f $env:JAVA_HOME, $env:PATH); \
 	Write-Host ('Updating PATH: {0}' -f $newPath); \
 # Nano Server does not have "[Environment]::SetEnvironmentVariable()"


tianon:
I'm not sure I've ever seen anything official using a `$java` environment variable (whereas `$JAVA_HOME` is reasonably standardized in my experience) -- do you have some documentation you could point to?

brunomartinspro:
> I'm not sure I've ever seen anything official using a `$java` environment variable (whereas `$JAVA_HOME` is reasonably standardized in my experience) -- do you have some documentation you could point to?

The documentation i based was on an issue of Azure DevOps, https://community.sonarsource.com/t/azure-devops-hosted-build-java-demanded-capability-not-present/8878/5

Manually creating a docker file for the [Azure DevOps agent](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/docker?view=azure-devops)
```
# Use the latest Windows Server Core image with .NET Framework 4.8.
FROM mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2019

#Install Java
FROM openjdk:14-ea-22-jdk-windowsservercore-1809

# Restore the default Windows shell for correct batch processing.
SHELL ["cmd", "/S", "/C"]

# Default to PowerShell if no other command specified.
CMD ["powershell.exe", "-NoLogo", "-ExecutionPolicy", "Bypass"]

#Create Working directoy
WORKDIR /azp

#Start Agent
COPY start.ps1 .

CMD powershell .\start.ps1

```
When the agent is running it doesn't have the Java capabilities, but adding an environment variable with the java complete path as "java" seems to work.

tianon:
I can appreciate that it fixed things in that case, but I don't think that's really a "standard" fix -- it sounds very particular to the expectations of that particular build job/script/environment, so I'm still not convinced we should add it to the images by default (especially since it's pretty simple for users to add downstream).

brunomartinspro:
I have to agree that this is not standard, will open an issue in Azure DevOps.

PR #384, opened 2020-01-10
Add support for windowsservercore-ltsc2019

aaronlahey:


tianon:
You might have already discovered this (hence the close), but just FYI / for posterity, `ltsc2019` is a rolling tag which is currently the same as `1809` (so we have opted not to explicitly build any `FROM ...:ltsc2019` images, preferring instead to follow the individual kernel compatibility tracks directly).

PR #390, opened 2020-02-05
Update remaining images from stretch to buster

### LINKED ISSUE #389: Rebuild 8-jdk image with Debian 10.0 ( buster )
The published openjdk image with tag **8-jdk** is having older debian image. Can we rebuild with latest image?

```
$ docker run -ti --entrypoint "/bin/sh" debian:buster-slim
# cat /etc/debian_version
10.2
# exit

$ docker run -ti --entrypoint "/bin/sh" openjdk:8-jdk
# cat /etc/debian_version
9.11
# exit
$
```

```
 $ trivy openjdk:8-jdk
2020-02-04T22:48:15.477+0530	INFO	Detecting Debian vulnerabilities...

openjdk:8-jdk (debian 9.11)
===========================
Total: 341 (UNKNOWN: 4, LOW: 46, MEDIUM: 227, HIGH: 55, CRITICAL: 9)
```

cc: @tianon @yosifkit

tianon:
Closes #389

(Not sure how/why we got a mismatch between `slim` and non-`slim` variants here, but this updates them all.)

### DIFF
--- a/11/jdk/Dockerfile
+++ b/11/jdk/Dockerfile
@@ -1,4 +1,4 @@
-FROM buildpack-deps:stretch-scm
+FROM buildpack-deps:buster-scm
 
 RUN set -eux; \
 	apt-get update; \
--- a/11/jre/Dockerfile
+++ b/11/jre/Dockerfile
@@ -1,4 +1,4 @@
-FROM buildpack-deps:stretch-curl
+FROM buildpack-deps:buster-curl
 
 RUN set -eux; \
 	apt-get update; \
--- a/8/jdk/Dockerfile
+++ b/8/jdk/Dockerfile
@@ -1,4 +1,4 @@
-FROM buildpack-deps:stretch-scm
+FROM buildpack-deps:buster-scm
 
 RUN set -eux; \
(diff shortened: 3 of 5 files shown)


tnir:
@tianon Thanks for fixing this. Affects me and I just spent hours to resolve this today 😄 

> (Not sure how/why we got a mismatch between slim and non-slim variants here, but this updates them all.)

Thanks again for addressing this. The root cause would be #342.

mazorga:
Hi @yosifkit,

Just wanted to mention that this change resulted in a huge jump in cpu consumption and a memory leakage.

Not sure really why, but we our deployment is spring web application (war) run with jetty-runner version 9.4.12.RC2 wrapped with datadaog(0.41.0) agent.

using openjdk:8-stretch (instead of openjdk:8)  solved the issue.

tianon:
> Not sure really why, but we our deployment is spring web application (war) run with jetty-runner version 9.4.12.RC2 wrapped with datadaog(0.41.0) agent.

For us to do anything about it, figuring out why would be the first step 😅

Even a minimal reproducer would help 👍

PR #401, opened 2020-04-01
Add Project Loom Build 15-loom+4-55

tmaret:
https://jdk.java.net/loom

### DIFF
--- a/15-loom/jdk/oracle/Dockerfile
+++ b/15-loom/jdk/oracle/Dockerfile
@@ -1,0 +1,57 @@
+FROM oraclelinux:7-slim
+
+RUN set -eux; \
+	yum install -y \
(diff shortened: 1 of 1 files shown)


tianon:
While Project Loom looks interesting, I'm not certain this is the right home for experimental pre-release Docker images of it (similar story for Project Metropolis, Project Panama, and Project Valhalla).

tmaret:
Thanks @tianon I'll keep it somewhere else.

delabassee:
I wouldn't publish such an image publicly without any big :warning: as Loom is EA and hence evolving rapidly! Besides, those builds are based on an **incomplete* JDK version.

Clearly and at this stage, it's best to source such project directly from https://jdk.java.net/

PR #407, opened 2020-04-28
Add initial GitHub Actions CI

tianon:
See https://github.com/docker-library/bashbrew/pull/2 (plus https://github.com/docker-library/official-images/pull/7888). :+1:


sormuras:
56 checks, nice matrix. 👍

tianon:
@sormuras it's really a little smaller than that -- it's doubled because of `push` vs `pull_request` (because I had to make the initial branch on the origin for it to even run the checks), but this does a test on every published tag based specifically on what's published to https://github.com/docker-library/official-images (instead of the "manual" matrix we'd maintained previously)

PR #412, opened 2020-05-12
Add initial Oracle arm64v8 support

tianon:
Oracle has added "aarch64" builds to https://jdk.java.net/15/ which I'm hoping will last past the EA period!


tianon:
Rebased and updated the generated code to include additional whitespace so it's more readable.

PR #423, opened 2020-07-20
Update Alpine to 3.12

tianon:


tianon:
Both Alpine builds are done and green. :+1:

PR #425, opened 2020-07-24
Update to Oracle Linux 8

tianon:
For `openjdk:14-jdk-oracle` this results in a ~9MB size increase.

cc @Djelibeybi -- do you think this is reasonable?  (any thoughts/review appreciated)

<details>
<summary>Diff:</summary>

```diff
$ diff -u <(bashbrew cat openjdk) <(bashbrew cat <(./generate-stackbrew-library.sh))
--- /dev/fd/63	2020-07-24 10:03:53.067812120 -0700
+++ /dev/fd/62	2020-07-24 10:03:53.071812001 -0700
@@ -1,10 +1,10 @@
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon), Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/openjdk.git
 
-Tags: 16-ea-7-jdk-oraclelinux7, 16-ea-7-oraclelinux7, 16-ea-jdk-oraclelinux7, 16-ea-oraclelinux7, 16-jdk-oraclelinux7, 16-oraclelinux7, 16-ea-7-jdk-oracle, 16-ea-7-oracle, 16-ea-jdk-oracle, 16-ea-oracle, 16-jdk-oracle, 16-oracle
+Tags: 16-ea-7-jdk-oraclelinux8, 16-ea-7-oraclelinux8, 16-ea-jdk-oraclelinux8, 16-ea-oraclelinux8, 16-jdk-oraclelinux8, 16-oraclelinux8, 16-ea-7-jdk-oracle, 16-ea-7-oracle, 16-ea-jdk-oracle, 16-ea-oracle, 16-jdk-oracle, 16-oracle
 SharedTags: 16-ea-7-jdk, 16-ea-7, 16-ea-jdk, 16-ea, 16-jdk, 16
 Architectures: amd64, arm64v8
-GitCommit: b64d9fd4cfc27a0a59d166ed4392dc5e984eb303
+GitCommit: e68ee469a61217d227514e7dd82dc3593bcb16de
 Directory: 16/jdk/oracle
 
 Tags: 16-ea-7-jdk-buster, 16-ea-7-buster, 16-ea-jdk-buster, 16-ea-buster, 16-jdk-buster, 16-buster
@@ -42,10 +42,10 @@
 Directory: 16/jdk/windows/nanoserver-1809
 Constraints: nanoserver-1809, windowsservercore-1809
 
-Tags: 15-ea-33-jdk-oraclelinux7, 15-ea-33-oraclelinux7, 15-ea-jdk-oraclelinux7, 15-ea-oraclelinux7, 15-jdk-oraclelinux7, 15-oraclelinux7, 15-ea-33-jdk-oracle, 15-ea-33-oracle, 15-ea-jdk-oracle, 15-ea-oracle, 15-jdk-oracle, 15-oracle
+Tags: 15-ea-33-jdk-oraclelinux8, 15-ea-33-oraclelinux8, 15-ea-jdk-oraclelinux8, 15-ea-oraclelinux8, 15-jdk-oraclelinux8, 15-oraclelinux8, 15-ea-33-jdk-oracle, 15-ea-33-oracle, 15-ea-jdk-oracle, 15-ea-oracle, 15-jdk-oracle, 15-oracle
 SharedTags: 15-ea-33-jdk, 15-ea-33, 15-ea-jdk, 15-ea, 15-jdk, 15
 Architectures: amd64, arm64v8
-GitCommit: 55eaa15885a2dd8b0d74c95fe9af71d90492b2d8
+GitCommit: e68ee469a61217d227514e7dd82dc3593bcb16de
 Directory: 15/jdk/oracle
 
 Tags: 15-ea-33-jdk-buster, 15-ea-33-buster, 15-ea-jdk-buster, 15-ea-buster, 15-jdk-buster, 15-buster
@@ -83,9 +83,9 @@
 Directory: 15/jdk/windows/nanoserver-1809
 Constraints: nanoserver-1809, windowsservercore-1809
 
-Tags: 14.0.2-jdk-oraclelinux7, 14.0.2-oraclelinux7, 14.0-jdk-oraclelinux7, 14.0-oraclelinux7, 14-jdk-oraclelinux7, 14-oraclelinux7, jdk-oraclelinux7, oraclelinux7, 14.0.2-jdk-oracle, 14.0.2-oracle, 14.0-jdk-oracle, 14.0-oracle, 14-jdk-oracle, 14-oracle, jdk-oracle, oracle
+Tags: 14.0.2-jdk-oraclelinux8, 14.0.2-oraclelinux8, 14.0-jdk-oraclelinux8, 14.0-oraclelinux8, 14-jdk-oraclelinux8, 14-oraclelinux8, jdk-oraclelinux8, oraclelinux8, 14.0.2-jdk-oracle, 14.0.2-oracle, 14.0-jdk-oracle, 14.0-oracle, 14-jdk-oracle, 14-oracle, jdk-oracle, oracle
 SharedTags: 14.0.2-jdk, 14.0.2, 14.0-jdk, 14.0, 14-jdk, 14, jdk, latest
-GitCommit: 29c17de6c8c0df6304d0c80f569deea1d4b6a23e
+GitCommit: e68ee469a61217d227514e7dd82dc3593bcb16de
 Directory: 14/jdk/oracle
 
 Tags: 14.0.2-jdk-buster, 14.0.2-buster, 14.0-jdk-buster, 14.0-buster, 14-jdk-buster, 14-buster, jdk-buster, buster
```

</details>

### DIFF
--- a/14/jdk/oraclelinux8/Dockerfile
+++ b/14/jdk/oraclelinux8/Dockerfile
@@ -1,4 +1,4 @@
-FROM oraclelinux:placeholder-slim
+FROM oraclelinux:8-slim
 
 RUN set -eux; \
 	microdnf install \
@@ -17,21 +17,29 @@ RUN set -eux; \
 # Default to UTF-8 file.encoding
 ENV LANG en_US.UTF-8
 
-ENV JAVA_HOME placeholder
+ENV JAVA_HOME /usr/java/openjdk-14
 ENV PATH $JAVA_HOME/bin:$PATH
 
 # https://jdk.java.net/
 # >
 # > Java Development Kit builds, from Oracle
 # >
-ENV JAVA_VERSION placeholder
+ENV JAVA_VERSION 14.0.2
 
 RUN set -eux; \
 	\
 	objdump="$(command -v objdump)"; \
 	arch="$(objdump --file-headers "$objdump" | awk -F '[:,]+[[:space:]]+' '$1 == "architecture" { print $2 }')"; \
 # this "case" statement is generated via "update.sh"
-	%%ARCH-CASE%%; \
+	case "$arch" in \
+# amd64
+		amd64 | i386:x86-64) \
+			downloadUrl=https://download.java.net/java/GA/jdk14.0.2/205943a0976c4ed48cb16f1043c5c647/12/GPL/openjdk-14.0.2_linux-x64_bin.tar.gz; \
+			downloadSha256=91310200f072045dc6cef2c8c23e7e6387b37c46e9de49623ce0fa461a24623d; \
+			;; \
+# fallback
+		*) echo >&2 "error: unsupported architecture: '$arch'"; exit 1 ;; \
+	esac; \
 	\
 	curl -fL -o openjdk.tgz "$downloadUrl"; \
 	echo "$downloadSha256 *openjdk.tgz" | sha256sum --strict --check -; \
--- a/15/jdk/oraclelinux8/Dockerfile
+++ b/15/jdk/oraclelinux8/Dockerfile
@@ -1,4 +1,4 @@
-FROM oraclelinux:placeholder-slim
+FROM oraclelinux:8-slim
 
 RUN set -eux; \
 	microdnf install \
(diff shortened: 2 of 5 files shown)


tianon:
I guess the `yum` -> `microdnf` breaking change is probably the most relevant thing to note here. :disappointed:

yosifkit:
Since users (like [us](https://github.com/docker-library/tomcat/blob/44d5a8ba6e7b85fd4d038e063456c1ae5d6a797a/10.0/jdk14/openjdk-oracle/Dockerfile)) could likely be installing packages, maybe not apply this to `14` since it is a GA release? Or maybe *more* variants... :cry:

tianon:
Who wrote that garbage `Dockerfile`????
(for readers not familiar: it me -- this is self-deprecation)

Begrudgingly good point. :sweat_smile:  I'll look at revamping the way this repo handles variants/base images first (so that it's more in line with how we maintain others like `python`, `php`, `golang`, etc).

Djelibeybi:
> **Edit:** no idea what I'm going to do with that `yumdb` stuff for `microdnf`..... does `microdnf` do/use something similar enough to even adapt that??

`microdnf` doesn't, but `dnf` does (you can `dnf mark` things). This doesn't solve your problem though. There are things you're doing downstream in that Tomcat `Dockerfile` that `microdnf` doesn't support (like `autoremove`).

I would stick with using `oraclelinux:7-slim` to be honest: it remains in production support until 2024.

tianon:
As tempting as it is to stay on Oracle Linux 7 and just ignore 8, I don't think that's very fair to our users (and we desperately needed a refactor anyhow).

I've now rebased this against https://github.com/docker-library/openjdk/pull/427, so this adds new `-oraclelinux8` variants, updates `-oracle` to be an alias of them, but leaves `-orcalelinux7` variants for existing users of the existing versions (with the intention that OpenJDK 17 will not support `-oraclelinux7` anymore).

Djelibeybi:
Fair enough. 

In which case, I have minor (non-functional) notes like:
* `microdnf` does not requiring the `-y` parameter) 
* `microdnf clean all` is actually useful so no need for the more heavy-handed `rm -rf /var/cache/yum`.

But both of those are stylistic choices and not even a hill I'm prepared to walk up, let alone get mildly injured on.

tianon:
Hey, I'm absolutely on board with minor stylistic changes to use the tools more "correctly" (especially in the case of things like `yum`, `dnf`, and `microdnf` where my own experience is lacking)! :metal:

Issue #493, opened 2022-04-20
ETA on April 2022 update for OpenJDK?


nobletrout:
Please see [this security advisory](https://mail.openjdk.java.net/pipermail/vuln-announce/2022-April/000015.html) from OpenJDK.

Multiple versions of openjdk have vulnerabilities that are patched in the latest JDK:
```
CVE-2022-21426
CVE-2022-21434
CVE-2022-21443
CVE-2022-21449
CVE-2022-21476
CVE-2022-21496
```

yosifkit:
The bot picked up the `18.0.1` release in https://github.com/docker-library/openjdk/commit/8a2f76391892edf7552ac3cb3273573d935bdc7c, but I don't see any others yet. Once the rest are available via their respective upstream location, they will be updated by the bot (see `versions.sh` for where it scrapes them).

For the process to get them to Docker Hub, see the [readme](https://github.com/docker-library/openjdk/tree/8a2f76391892edf7552ac3cb3273573d935bdc7c#see-a-change-merged-here-that-doesnt-show-up-on-docker-hub-yet).

skagedal:
So, the script scrapes the page https://jdk.java.net/17/ and that still lists 17.0.2. So while this might be offtopic for this forum – any ideas on why this might be, and if there would be some channel to ping to make Oracle update it? [Bug report](https://bugreport.java.com/bugreport/)?

areguru:
> So, the script scrapes the page https://jdk.java.net/17/ and that still lists 17.0.2. So while this might be off topic for this forum – any ideas on why this might be, and if there would be some channel to ping to make Oracle update it? [Bug report](https://bugreport.java.com/bugreport/)?
> 
> [edit: I did send a bug report through that channel, don't know if that will help but can't hurt... Also tried the e-mail address listed at the above linked page "for difficulty downloading", but that just bounced.]

@skagedal I sent a complaint about https://jdk.java.net/17/ linking only to 17.0.2 and not 17.0.3 to "Contact us" on the page 
https://www.oracle.com/java/technologies/downloads/

areguru:
Hi again

I got the following answer from Oracle:

> Currently, Oracle only publishes the first two updates to a major JDK version (e.g., 17.0.1, 17.0.2).  Control of the version is then turned over to the community.  The JDK 17 Updates project[0] handles patches, etc. from that point on.  There's also a downloads section[0] that gives a recommendation for a public build of the JDK.
> 
[...]

PR #495, opened 2022-04-26
Remove OpenJDK 17

tianon:
There no longer exist "official" (or even semi-official) vanilla builds of OpenJDK 17 suitable for our use or for publishing as "OpenJDK" (https://jdk.java.net/17/).

https://github.com/docker-library/openjdk/issues/493#issuecomment-1109591179


tianon:
FYI to affected maintainers:

- `clojure`: @Quantisan @cap10morgan
- `jetty`: @gregw @lachlan-roberts @olamy @joakime
- `jruby`: @jruby @headius @enebo
- `maven`: @carlossg
- `tomcat`: @tianon @yosifkit
- `tomee`: @lordofthejars @otaviojava @jgallimore @scriptmonkey

I would also suggest taking a look at https://github.com/docker-library/docs/pull/2142 (which explains better why this isn't something we really can fix for this repository :disappointed:).

cap10morgan:
Thanks for the heads-up, @tianon! So it sounds like we downstream maintainers need to select one of the other images to base ours on? What a mess (looking at you, Oracle).

joakime:
Thanks for creating this issue.

It will also help on our side when we eventually get the "what happened to openjdk-17 images?" questions.
We can now point them here.

headius:
Ugh, ok. So it looks like we should be able to switch to coretto or temurin without any major impact. I expect there's going to be concerns from downstream users but I can point them back here.

headius:
To clarify... this only affects 17+ correct? It will be a much less risky change if I can leave 11 and 8 on openjdk images.

nicolas-albert:
Just in time before our major release 👍 
I switch convertigo base to tomcat:9-jdk17-temurin.
We already use Temurin for our Eclipse based studio without issues.
Thanks for the mention !

tianon:
> To clarify... this only affects 17+ correct? It will be a much less risky change if I can leave 11 and 8 on openjdk images.

Yeah, that's correct - just the builds provided by Oracle (which is anything other than 8 and 11).  However, as you might've seen with the vanilla builds of 8 and 11, they're not updated on an extreme priority (understandably, given they're vanilla builds provided with no expectation of support), so maybe just keep that in mind. :slightly_smiling_face:

cap10morgan:
@tianon Are [these are the security vulnerabilities in 17.0.2](https://www.oracle.com/security-alerts/cpuapr2022.html#AppendixJAVA) that are motivating a quick push to 17.0.3?

tianon:
@cap10morgan I'm not sure what you mean in this context :sweat_smile:

I do believe that 17.0.3 contains security updates, but this PR (and the `openjdk` official image) will not be receiving 17.0.3. :grimacing: :see_no_evil:

cap10morgan:
@tianon Yeah I get that openjdk won't get 17.0.3. But it seemed like there was some urgency to update to 17.0.3 (more than just the usual "oh there's a new patch release out"), so just wanted to clarify if this was the source of the urgency. But perhaps I misunderstood.

scriptmonkey:
> FYI to affected maintainers:
>     * `tomee`: @lordofthejars @otaviojava @jgallimore @scriptmonkey


Thank you for tagging us in this PR!  It would seem that TomEE will be moving to Temurin.

syphr42:
@cap10morgan CVE-2022-21449 might be a good reason to make 17.0.3 available soon, but it's unclear to me if it affects only Oracle builds.

cap10morgan:
@syphr42 I don't think they _do_ only affect Oracle builds. It's just that Oracle is alone in not releasing a fixed version.

cap10morgan:
> @tianon Yeah I get that openjdk won't get 17.0.3. But it seemed like there was some urgency to update to 17.0.3 (more than just the usual "oh there's a new patch release out"), so just wanted to clarify if this was the source of the urgency. But perhaps I misunderstood.

Oh, it was the description of https://github.com/docker-library/openjdk/issues/493 that sent me down this "it's motivated by fixing some security vulns" rabbit hole. Sorry, should have commented over there. :)

PR #532, opened 2023-10-11
Increase default RAM percentage for JVM

tariqajyusuf:
The JVM by default only uses up to 25% of available memory. Since containers are running only one JVM it makes sense to bump this up to a higher amount to better use the total container space.


LaurentGoderre:
Thanks for the contribution. However, as per the https://github.com/docker-library/openjdk/blob/master/README.md this image is now deprecated.

yosifkit:
We only provide pre-release OpenJDK images from this repo. We also try to not stray from upstream default values.

PR #533, opened 2023-12-09
Add 23-ea+1

marchof:
Hi, it's me from javaalmanac.io again 👋

As we have a first Java 23 EA build I tried to add it here following the pattern in commit 323e8d8de1ae7e3cfaeb85e20551a5e895461296.

Thanks for maintaining these useful docker images!

### DIFF
--- a/23/jdk/bookworm/Dockerfile
+++ b/23/jdk/bookworm/Dockerfile
@@ -1,0 +1,96 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "apply-templates.sh"
+#
+# PLEASE DO NOT EDIT IT DIRECTLY.
+#
+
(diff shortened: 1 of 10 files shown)


marchof:
@tianon Is there anything else that needs to be done to make this version show up at hub.docker.com?

tianon:
Can you be more specific? :sweat_smile:

As far as I can tell, they went up successfully: https://hub.docker.com/_/openjdk/tags?name=23

(https://github.com/docker-library/openjdk#see-a-change-merged-here-that-doesnt-show-up-on-docker-hub-yet -> via https://github.com/docker-library/official-images/pull/15890)

marchof:
@tianon My bad, it was another problem in my pipeline. It's up and running now: https://javaalmanac.io/jdk/23/

PR #535, opened 2024-03-14
Add oraclelinux9 variants

tianon:


yosifkit:
Are there template updates too?

tianon:
Yep!

yosifkit (approved):

PR #537, opened 2024-05-29
Ask RPM for its own arch instead of objdump hacks

tianon:
This is much cleaner and should be more appropriately granular / generally reliable.

(With special thanks to @grooverdan again :heart:)


tianon:
(@gdams @sxa y'all might also be interested in this, as I believe you've copied my `objdump` hacks into `eclipse-temurin`?)

tianon:
(@jennyowen @glindroth @ericsporre y'all have `uname -m` in `neo4j` and it might be worth replacing with something like this :eyes:)

PR #544, opened 2025-01-22
Add Windows Server 2025 variant

yosifkit:
Add Windows Server ltsc2025; keep 1809 as it is still in extended support (we may drop it later, even while it is supported by Windows).

Refs: https://github.com/docker-library/official-images/pull/18143

### DIFF
--- a/versions.sh
+++ b/versions.sh
@@ -166,6 +166,7 @@ for version in "${versions[@]}"; do
 				| "alpine" + . else empty end,
 				if $doc.jdk.arches | keys | any(startswith("windows-")) then
 					(
+						"ltsc2025",
 						"ltsc2022",
 						"1809",
 						empty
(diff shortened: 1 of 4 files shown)


versions.sh
```diff
@@ -166,6 +166,7 @@ for version in "${versions[@]}"; do
 				| "alpine" + . else empty end,
 				if $doc.jdk.arches | keys | any(startswith("windows-")) then
 					(
+						"ltsc2025",
```
tianon, line 169: It's interesting and mostly orthogonal that we didn't add any Nano Server past 1809 here -- something worth digging into later, maybe.

tianon (approved):

versions.sh
tianon: Maybe it's something like https://github.com/docker-library/mongo/pull/720#issuecomment-2606055808 :grimacing: :see_no_evil:

tianon: After we merge this, maybe we can try adding them back again so we can add useful comments.

PR #545, opened 2025-01-22
Add more Nano Server variants

tianon:
This is a trial (see https://github.com/docker-library/openjdk/pull/544#discussion_r1924547165).

If it works, we can merge it as-is, if it doesn't, we can update the code to comment these out with a note about what failed.


tianon:
After digging a little, it looks like this was an honest oversight in https://github.com/docker-library/openjdk/pull/472 -- the tests all pass (and we run two separate Java tests on these images). :shrug:

yosifkit (approved):

PR #550, opened 2025-10-31
Remove misleading "bare" aliases

tianon:
These make it seem like `openjdk:26` is GA, and we don't want that (in fact, we aggressively don't want that).

This also removes a lot of old cruft we haven't supported/used in ~3 years, such as Alpine support (last supported on jdk.java.net in 2022) and JRE support (last supported here in 2022 as well, with prior builds).

My ultimate intent is to delete a _lot_ of tagged content from Docker Hub shortly following this (basically anything that isn't obviously either an old explicit historical version like "11.0.6" or clearly Early Access/pre-release like "14-ea" or "21-rc").


tianon:
Deletes are now complete (~6.5% of the total 16,288 tags `openjdk` had before this).

JPRoyer:
docker.io/library/openjdk:8-jdk-alpine: not found too here

colinmoynes:
> > My ultimate intent is to delete a _lot_ of tagged content from Docker Hub shortly following this (basically anything that isn't obviously either an old explicit historical version like "11.0.6" or clearly Early Access/pre-release like "14-ea" or "21-rc").
> 
> Explicit diff of tags I intend to remove/delete from Docker Hub (to encourage users to check the image description and go elsewhere or be more explicit about their opt-in to deprecated/unsupported/unmaintained/historical content):

This seems a little ill thought out with regards to users using these images in production. The sudden removal of them has caused problems. I get the intentions, but really, what purpose did it serve?

